PR #42, opened 2015-01-24
diff cap

danielnaber:
Panorama sometimes jumps after inertia. This seems to fix it. How to reproduce:

* pan fast so that inertia get active
* wait until it doesn't pan from inertia anymore
* wait another 2-3 seconds
* pan fast again -> panorama jumps


mpetroff:
Thanks. Your pull request had a bunch of extra commits, so I applied the patch using the terminal.

PR #44, opened 2015-01-29
Update pannellum.js

umbreak:
Added a JSON config option to stop autorotate after a certain time (counting from when the autorotate started).

The name of the configuration option is `autoRotateStopAfter` and by default is inactive.

That can be useful for when we want the image to autorotate in a interval (between x and y).


mpetroff:
Thank!

mpetroff:
I renamed the configuration option to `autoRotateStopDelay`.

PR #49, opened 2015-02-06
Update pannellum.js

umbreak:
Adding to JSON config file `hotspotImage`. This property defines the hotspot image URL.
* I deleted the image starting size (in %), cause it makes more sense to define it only in the CSS.

### DIFF
--- a/src/js/pannellum.js
+++ b/src/js/pannellum.js
@@ -67,6 +67,7 @@ var defaultConfig = {
     vOffset: 0,
     autoRotate: false,
     autoRotateInactivityDelay: -1,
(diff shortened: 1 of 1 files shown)


mpetroff:
I'm not going to merge this or #50. This doesn't add anything that wasn't already possible with custom CSS, and it's poorly implemented. If I were to add this feature, which I'm not sure I even want included, it would be configurable for each individual hotspot and would not be limited to a specific hotspot type. Additionally, you haven't followed the existing coding style, e.g. your use of tabs, and your commit messages and pull request titles are incredibly uninformative.

PR #50, opened 2015-02-06
Update pannellum.css

umbreak:
This is an example where the image is set by default to 70% size, and the hover to 100% (using `CSS transitions`)
There is also an example of `CSS Keyframes animation` commented.

The right way to do this (not hardcoded) would be to **provide a [LESS](http://lesscss.org/) file, with the `initialWidth, hoverWidth and animationKeyFrameName` as variables**. 
Then we could provide diferent animation keyframes (which is used for the variable animationKeyFrameName) in the  in a separate css file. For example a fade animation, a width animatino, a change_color animation.

### DIFF
--- a/src/css/pannellum.css
+++ b/src/css/pannellum.css
@@ -396,3 +396,85 @@ div.tooltip:hover:after {
 #lmsg {
     font-size: 12px;
 }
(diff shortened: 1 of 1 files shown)


mpetroff:
Image bulger is unwanted. Also see comment on #49.

PR #54, opened 2015-02-19
Extra space not matching the replace condition on build

umbreak:
Hello,

the minified built version of pannellum.htm was not correct cause this space in the replace line was not replacing the corresponding line in src/pannellum.htm

        <link type="text/css" rel="Stylesheet" href="css/pannellum.css"/>


mpetroff:
Thanks!

PR #58, opened 2015-02-27
Pitch yaw parameters

hhkaos:
I wasn't able to find out how to know the pitch and yaw of a point on an image so I made this small example to solve it hehe. I'm not sure which is the best way to do it but ... this.. at least works ;D

### DIFF
--- a/src/js/pannellum.js
+++ b/src/js/pannellum.js
@@ -478,6 +478,12 @@ function onDocumentMouseMove(event) {
         var pitch = ((Math.atan(event.clientY / renderer.canvas.height * 2 - 1) - Math.atan(onPointerDownPointerY / renderer.canvas.height * 2 - 1)) * 180 / Math.PI * vfov / 90) + onPointerDownPitch;
         pitchSpeed = (pitch - config.pitch) * 0.2;
         config.pitch = pitch;
+
+        var verbose = config.verbose || false;
+        if(verbose){
+            console.log("Yaw = ", yaw);
+            console.log("Pitch = ", pitch);
+        }
     }
 }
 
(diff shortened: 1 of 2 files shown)


src/js/pannellum.js
```diff
@@ -478,6 +478,12 @@ function onDocumentMouseMove(event) {
         var pitch = ((Math.atan(event.clientY / renderer.canvas.height * 2 - 1) - Math.atan(onPointerDownPointerY / renderer.canvas.height * 2 - 1)) * 180 / Math.PI * vfov / 90) + onPointerDownPitch;
         pitchSpeed = (pitch - config.pitch) * 0.2;
         config.pitch = pitch;
+
+        var verbose = config.verbose || false;
+        if(verbose){
+            console.log("Yaw = ", yaw*-1);
```
gautiermichelin: Hi hhkaos, just a single question, why that -1 here ?

hhkaos: Hehe, good question, I'm not sure why either, but if I take the original yaw error it displays the point on a different position. Is it that possible?

mpetroff: That would be a bug in the hot spot rendering. I just fixed it, which unfortunately means the sign needs to be switched for the yaw of every hot spot in existing configuration files. It's rather embarrassing that I never noticed this.

mpetroff:
This is an excellent idea, although I disagree with the implementation. I think it would be better to log the pitch and yaw values of mouse clicks instead. I'll try to add that in the next few days. See #59.

mpetroff:
Implemented in f2a787fdbc282ee68df53bd38af1c2199aaeb3f6.

PR #67, opened 2015-03-17
Fix 'undefined' path elements if using multiRes without explicit setting it's basePath config value + minor improvements

mko-x:
As multiRes.basePath is probably an optional value to set, it should work without setting as well. Before this commit the missing multiRes.basePath lead to pathelements à la 'undefined/' or 'undefined.


mpetroff:
Thanks!

mko-x:
You're welcome!

I really like your Pannellum and you can expect several more pull requests :)

PR #70, opened 2015-04-13
Added an option for specifying the multires cube size

jlehtine:
Being able to specify the cube size manually for the multires generation script using -c/--cubesize enables generation of compact multires presentations from huge panoramas and the use of power-of-two cube and tile sizes. I have found cube size 2048 to be especially useful given typical display resolutions and hfov. Default is to determine cube size automatically as before.


mpetroff:
Thanks!

PR #72, opened 2015-05-05
Fix undefined error when moving too fast

umbreak:
If we are switching from one image to another too fast, oldRenderer is undefined at that point. This fix that.


mpetroff:
See #73.

PR #74, opened 2015-05-11
Moving from visibility to display due to Chrome bug

umbreak:
If you animate the hotspot (which is not the default case for pannellum, but which it could be) with an infinite CSS keyFrame animation, `vissibility:hidden` has an error in Chrome, as I described here: http://stackoverflow.com/questions/30167467/css-infinite-animation-after-hidden-is-not-resetted-chrome

### DIFF
--- a/src/css/pannellum.css
+++ b/src/css/pannellum.css
@@ -282,7 +282,7 @@ body {
     position: absolute;
     height: 26px;
     width: 26px;
-    visibility: hidden;
+    display: none;
     cursor: default;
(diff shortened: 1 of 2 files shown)


mpetroff:
It turns out that this broke the display of Pannellum's default tooltips, so I reverted it. I'll be happy to merge something back in if it doesn't break the defaults.

umbreak:
Well, then we have to wait until they solve the issue: https://code.google.com/p/chromium/issues/detail?id=444852

PR #84, opened 2015-08-06
CSS and events changes for better embed viewer

Benvii:
Hi,
I tried to embed the viewer on an existing webpage (not using an iframe) and noticed some CSS problems and also events that were listening on document instead of the viewer container.

Here are some changes that fixes these problems.

About the standalone (iframed) version I added a standalone.css file to preserve the styles applied on html and body that were removed from pannellum.css as they should only be applied to the container.
I decided to include this stylesheet in the compiled version as standalone.js is included, but I don't think it's a good choice because the compiled version shouldn't apply style to html or even body elements.

Maybe standalone.js and standalone.css should be directly included into standalone.htm and separated from pannellum.js and pannellum.css, so that the compiled version can be used in embeded mode (not iframed mode).
What's your opinion ?

Thanks for this great viewer, I will also try to start an event handler, and create specific events tomorrow, for instance an event to know when the viewer is loaded, when pitch/yaw change ....


mpetroff:
Thanks! I hadn't tried using the library version yet; so far I was just making sure I didn't introduce any regressions in the standalone version.

I didn't mean to include `standalone.js` is the minified `pannellum.js`. I added `standalone.js` to the build process when it was still only producing `pannellum.htm`. When I went back later and modified the build process to produce `pannellum.js`, it didn't cross my mind that I was still including `standalone.js`. I just fixed it.

PR #86, opened 2015-08-14
Fix 'undefined' bug in fallback URLs (in library version)

Benvii:
Fixes a small bug noticed in library version.
basePath isn't checked for fallback tiles, so undefined appear in the URL fallback tile URL if basePath isn't defined.


mpetroff:
Thanks!

PR #120, opened 2016-03-08
enable generates.py to take cube faces as input

franckinux:
Hello,
I've modified the generates.py file in order to take cube faces as input.
my workflow is as follow : after generating the cubes faces from the equirectangular image, I edit the down face and then I make tiles from the cube faces. I usually delete the equirectangular image since it becomes useless.
as I found it inadequate to build up the equirectangular image from the cube faces to comply with generates.py's input, I decied to modify it to accept cube faces as input.
if the input is a directory, it must contain 6 files whose first letters are in "fbudlr".
Hope it helps,
Franck

### DIFF
--- a/utils/multires/generate.py
+++ b/utils/multires/generate.py
@@ -4,17 +4,18 @@
 
 # generate.py - A multires tile set generator for Pannellum
 # Copyright (c) 2014-2015 Matthew Petroff
-# 
+# Copyright (c) 2016 Franck Barbenoire
+#
 # Permission is hereby granted, free of charge, to any person obtaining a copy
 # of this software and associated documentation files (the "Software"), to deal
 # in the Software without restriction, including without limitation the rights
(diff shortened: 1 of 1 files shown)


mpetroff:
I'd generally like to keep the script simple, so I'd rather not make it considerably longer by adding support for a specific workflow. The pull request looks good, but it's not really something I want to add.

franckinux:
Thank you for your answer,
I don't think this is a particular workflow because I edit my nadir (I guess most people do like this) in the down cube face after having converted the equirectangular image to cube faces.
I understand you want to keep generate.py simple, so what about a separate new python script ?
Franck

mpetroff:
I extract the nadir, edit it, and add it into the stitch project before I stitch the final equirectangular image. I have no idea which method it more common, though.

I would like to keep `generate.py` simple for easy of maintenance; having a separate script would be worse.

PR #124, opened 2016-03-13
hs.image has precedence on hs.URL

Cyrille37:
Hi,
I'd changed hs.image precedence on hs.URL to permits specifying a optional url on image click.
Cheers


mpetroff:
I made this change in 943d29d707edb891d606eb7ded10219736adef37 but didn't use your implementation. I also changed the code to make source URLs for image and video hot spots relative to the configuration file instead of `pannellum.htm` to be consistent with everything else.

PR #131, opened 2016-03-25
Device orientation API

mkbit:
I added the device orientation API to pannellum so a mobile phone user can view the panorama by tilting of the phone if the new button or the compass is clicked. Since the specification of W3C is still a draft I marked this feature as experimental.

So far I was able to test it with Chrome and Firefox on Android.
Firefox doesn't behave like the specification and the image suddenly rotates 180° at certain points in landscape mode. At the moment I am not able to investigate it any further since Firefox is blocking WebGL on my phone.

### DIFF
--- a/doc/json-config-parameters.md
+++ b/doc/json-config-parameters.md
@@ -118,6 +118,13 @@ viewer can be set to. Defaults to `50` / `120`.
 
 If `true`, a compass is displayed. Defaults to `false`.
 
+### `enableDeviceOrientationCtrl`
+
+*experimental*
+
+If `true`, the panorama can be viewed by panning of your device (only on devices that support the device orientation api).
+This feature can be enabled by clicking the button that appears if it is available or clicking the compass.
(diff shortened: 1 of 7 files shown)


mpetroff:
A proper implementation of this currently isn't possible since `libpannellum` doesn't currently support setting the camera roll (for the most part), which is necessary to handle device rotations. Roll support is currently part of the equirectangular fragment shader but isn't exposed; support needs to be added to the cubemap fragment shader and to the multires geometry calculations. As such, I'm going to leave this pull request unmerged but open instead of adding incomplete support; once I've added the required roll support, I'll revisit it.

mpetroff:
After adding rotation support in 08f4d8955bf289d7ed534e9b0a4fdd91c500dbc8 and 08f4d8955bf289d7ed534e9b0a4fdd91c500dbc8, I reimplemented this feature in b0c2f03511c09f1b140079b5cf99281bca46d023. Therefore, I'm closing this pull request.

PR #132, opened 2016-04-12
Update readme.md

saidmoya12:
Add link to web sample


mpetroff:
The NPM part is wrong since Pannellum isn't a CommonJS module, and there's no reason why the examples directory should be changed to a URL. I'm rejecting this pull request.

saidmoya12:
I installed it via npm without problems using require('pannellum')
Aditionally i'm working on npm full support becouse my project only work via npm modules

When I finished the implementation I will send a pull request if you wont

PR #135, opened 2016-04-18
Update pannellum.js

saidmoya12:
Resize on orientationchange event for resize.

This fix an issue on mobile devices.

### DIFF
--- a/src/js/pannellum.js
+++ b/src/js/pannellum.js
@@ -387,6 +387,7 @@ function onImageLoad() {
         container.addEventListener('msfullscreenchange', onFullScreenChange, false);
         container.addEventListener('fullscreenchange', onFullScreenChange, false);
         window.addEventListener('resize', onDocumentResize, false);
+		window.addEventListener('orientationchange', onDocumentResize, false);
         container.addEventListener('keydown', onDocumentKeyPress, false);
         container.addEventListener('keyup', onDocumentKeyUp, false);
         container.addEventListener('blur', clearKeys, false);


src/js/pannellum.js
```diff
@@ -387,6 +387,7 @@ function onImageLoad() {
         container.addEventListener('msfullscreenchange', onFullScreenChange, false);
         container.addEventListener('fullscreenchange', onFullScreenChange, false);
         window.addEventListener('resize', onDocumentResize, false);
+		window.addEventListener('orientationchange', onDocumentResize, false);
```
sukima: Indenting is four spaces.

sukima:
At first glance it is confusing just what *Update pannellum.js* means. Could be anything. It requires a larger degree of cognitive energy to dive into the code and understand the abstract message. Perhaps something like the following would provide a more descriptive commit message:

```
Resize with device orientation changes

When a mobile device changes orientation it does not trigger the
`resize` event like a normal window resize would. This means on mobile
devices that changing the orientation does not cause the pannellum
canvas to resize. This commit resizes based on the `orientationchange`
event.
```

You can fix these using the `git commit --amend` option.

sukima:
In fact [I did just that](https://gist.github.com/sukima/a75934a4900a6b1830891371bf45eb81):

```patch
From 7c7e8615860eb40da0e95ae41d9f750079ac9eb4 Mon Sep 17 00:00:00 2001
From: Said <saidmoya12@gmail.com>
Date: Mon, 18 Apr 2016 14:45:03 -0500
Subject: [PATCH] Resize with device orientation changes

When a mobile device changes orientation it does not trigger the
`resize` event like a normal window resize would. This means on mobile
devices that changing the orientation does not cause the pannellum
canvas to resize. This commit resizes based on the `orientationchange`
event.
---
 src/js/pannellum.js | 1 +
 1 file changed, 1 insertion(+)

diff --git a/src/js/pannellum.js b/src/js/pannellum.js
index c216432..48029f1 100644
--- a/src/js/pannellum.js
+++ b/src/js/pannellum.js
@@ -387,6 +387,7 @@ function onImageLoad() {
         container.addEventListener('msfullscreenchange', onFullScreenChange, false);
         container.addEventListener('fullscreenchange', onFullScreenChange, false);
         window.addEventListener('resize', onDocumentResize, false);
+        window.addEventListener('orientationchange', onDocumentResize, false);
         container.addEventListener('keydown', onDocumentKeyPress, false);
         container.addEventListener('keyup', onDocumentKeyUp, false);
         container.addEventListener('blur', clearKeys, false);
-- 
2.8.1

```

And it can be applied onto `master` with:

```console
$ curl https://gist.githubusercontent.com/sukima/a75934a4900a6b1830891371bf45eb81/raw/bc539564156c6fea47c9f6914a322e06fd9519cc/0001-Resize-with-device-orientation-changes.patch | git am
```

mpetroff:
I added this fix in 6966e57d14eac25e10aa3fecffcfc366a1dabe86. As @sukima mentioned, your indenting doesn't follow the style of the rest of the code and neither the commit message nor the pull request title are informative (however, I think a one line description would be plenty). Therefore, I didn't merge the pull request.

saidmoya12:
+1

Issue #155, opened 2016-05-17
Hotspots misplaced in fullscreen mode in Safari on OS X


mzoeller:
When switching to fullscreen mode in Safari (9.0.3) on OS X hotspots are misplaced and aspect ratio is skewed. Non fullscreen mode works fine. 

Example: http://www.panoheld.de/portfolio-item/ferienwohnung-studio44/

mpetroff:
I think this is fixed. Can you please test the fix?

mzoeller:
It is! Working fine now.

PR #156, opened 2016-05-28
Fix wrong aspect ration on OS X Safari

mzoeller:
This fixes https://github.com/mpetroff/pannellum/issues/155. I'm sure it can be done more elegantly but it should point you in the right direction, basically recalculate aspect ration after full screen was entered and render one frame so that screen does not stay black.


mpetroff:
This did point me in the correct direction. It seems the resize event wasn't firing. I think I fixed it in d1615f4b6a82081c9b16b08bc6cbfd0b2b84c878. Thanks.

PR #160, opened 2016-06-05
Enable additional URL configuration parameters

jannecederberg:
I would suggest adding support for the following parameters from URL config:
  - minHfov, maxHfov
  - minPitch, maxPitch
  - minYaw, maxYaw

The motivation for this was making it easier to create auto-generated
panoramas using Jekyll/Hugo and other static site generators without
needing to create JSON files for panoramas.

If a given panorama is not a full 360x180 panorama, then the above
parameters are needed in addition to haov and vaov parameters
in order for the user not to see black areas in the panorama.
Hence suggesting adding URL support.

### DIFF
--- a/doc/url-config-parameters.md
+++ b/doc/url-config-parameters.md
@@ -14,6 +14,7 @@ Specifies the URL of a JSON configuration file.
 ## Other parameters
 
 A subset of the JSON configuration file options can be used as URL parameters.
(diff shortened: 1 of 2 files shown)


mpetroff:
Thanks. I'm generally hesitant about adding about adding additional URL parameters, but you made a convincing argument with regards to static site generators.

jannecederberg:
Yes, I understand your point of view, that's why I wrote the justification for the idea/request. Thank you for merging!

PR #163, opened 2016-06-11
Introduce a dynamic coefficient for touchmove pan (yaw/pitch change) speed

jannecederberg:
With a fixed coefficient affecting touchmove pan speed, at large values of config.hfov (zoomed out) panning will feel "sluggish" due to panorama panning less than finger is moved on screen. On the other hand, at small config.hfov values (zoomed in), panning will be perceived to move faster.

To improve usability at both small and large config.hfov (zoomed-out and zoomed-in) we introduce a dynamic coefficient affected by the config.hfov value.

This seems to roughly keep initial drag/pan start position close to the user's finger while panning regardless of zoom level.

### DIFF
--- a/src/js/pannellum.js
+++ b/src/js/pannellum.js
@@ -736,12 +736,21 @@ function onDocumentTouchMove(event) {
             setHfov(config.hfov + (onPointerDownPointerDist - clientDist) * 0.1);
             onPointerDownPointerDist = clientDist;
         }
-        
-        var yaw = (onPointerDownPointerX - clientX) * 0.1 + onPointerDownYaw;
+
+        // The smaller the config.hfov value (the more zoomed-in the user is), the faster
+        // yaw/pitch are perceived to change on one-finger touchmove (panning) events and vice versa.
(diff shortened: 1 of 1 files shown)


jannecederberg:
Oh yeah, true, I guess I implicitly noticed it was implemented for mouse as I don't recall paying attention to it on the computer :)

The coefficient I used was just a quick-n-simple try and could definitely be improved upon; now looking at your mouse implementation I see you account for canvas dimensions/aspect ratio which I noticed through experimentation to seem to also affect perceived panning speed but ran out of time to implement it. Should've looked at your mouse coeff earlier :)

Thanks again for merging!

jannecederberg:
Oh yeah, true, I guess I implicitly noticed it was implemented for mouse as I don't recall paying attention to it on the computer :)

The coefficient I used was just a quick-n-simple try and could definitely be improved upon; now looking at your mouse implementation I see you account for canvas dimensions/aspect ratio which I noticed through experimentation to seem to also affect perceived panning speed but ran out of time to implement it. Should've looked at your mouse coeff earlier :)

Thanks again for merging!

PR #166, opened 2016-06-15
Prevent empty values in URL config from stopping panorama rendering

jannecederberg:
Should the user use URL configuration for their panorama and then specify an option/key but leave the value empty, previously this would stop the panorama from being displayed due to the generated JSON config from src/standalone/standalone.js being invalid.

This commit omits creating an on-the-fly JSON string and instead uses a JavaScript object (which the generated JSON string was converted to anyway).

The background of wanting to enable URL options with possibly empty values relates to using Pannellum with static site generators. Depending on the static site generator, specifying default values for undefined template variables can result in lots of template notation whereas outputting a value for a variable that isn't defined usuallly simply outputs an empty string. It would hence be desirable to treat the empty-string value of an option/key in URL config as if that option/key were not user-supplied at all and hence the Pannellum default value for that option/key be used.

### DIFF
--- a/doc/json-config-parameters.md
+++ b/doc/json-config-parameters.md
@@ -72,6 +72,10 @@ If set to `false`, the zoom controls will not be displayed. Defaults to `true`.
 
 If set to `false`, zooming with keyboard will be disabled. Defaults to `true`.
 
+### `mouseZoom`
+
+If set to `false`, zooming with mouse wheel will be disabled. Defaults to `true`.
+
 
 ### `showFullscreenCtrl`
 
--- a/src/js/pannellum.js
+++ b/src/js/pannellum.js
@@ -83,7 +83,8 @@ var defaultConfig = {
     northOffset: 0,
     showFullscreenCtrl: true,
     dynamic: false,
-    keyboardZoom: true
+    keyboardZoom: true,
+    mouseZoom: true
 };
 
 // Initialize container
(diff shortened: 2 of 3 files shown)


jannecederberg:
Hmm, I pushed another commit to jannecederberg/master but didn't mean for that commit to become part of this pull request but meant to create another pull request about it.

mpetroff:
Why was `autoRotate` moved to be with `autoLoad` and `ignoreGPanoXMP` instead of with the other parameters where it was?

mpetroff:
I applied the second commit. It's generally best to create [feature branches](https://guides.github.com/introduction/flow/) for each pull request so new commit don't get added to an existing pull request.

jannecederberg:
**About using feature branches:**
Ok, yes, apologies for not understanding the feature branch approach sooner! I'm kind of new to pull requests on GitHub. Thank you for accepting the mouseZoom part!

**About moving `autoRotate` to be with `autoLoad` and `ignoreGPanoXMP` as parsed input:**
Based on documentation I interpreted `autoRotate` to expect a float but based on code review I came to the conclusion that it's a mixed variable, either boolean or float. If/when using the JavaScript config object approach in src/standalone/standalone.js that I suggested (instead of constructing a JSON string and then parsing it using `JSON.parse`), if `autoRotate` is then not parsed but taken as is and a boolean value (as string) is specified by the user (via URL config), then a string containing "true" / "false" will be set as `autoRotate` value and the resulting panorama (at least according to my quick tests) will look along the lines of this image: <http://imgur.com/lFA6dzP>

Hence I moved `autoRotate` to be parsed as JSON.

jannecederberg:
The current approach will result in `hfov`, `pitch`, `yaw`, `haov`, `vaov`, `vOffset`, `minHfov`, `maxHfov`, `minPitch`, `maxPitch`, `minYaw` and `maxYaw` being set as string values containing integers/floats. Maybe using `JSON.parse()` for them as well or `parseFloat()` ?

My background motivation/justification for changing from `JSON.parse(string json)` to using a JavaScript object was that I felt it was more pleasant omitting URL params with empty values when using a JavaScript object than when dealing with JSON's "list cannot end in a comma" stuff and then parsing the JSON string :)

mpetroff:
The `autoRotate` parameter is supposed to be a number, although it's set to `false` in certain places in the code. It happens to work when set to `true` because JavaScript automatically casts `true` to `1` when performing an operation with a number (a string can't be cast, so the operation results in a `NaN` value, hence the display bug).

I moved the `autoRotate` parameter back with the numbers, added an explicit string-to-number conversion, and committed the changes in b1bbc8622420a040bc94038100fa7fd78391ad30.

Thanks for both changes.

jannecederberg:
Ok, great, thank you and welcome!

Issue #92, opened 2015-10-09
minPitch and maxPitch calculation should take zoom level as a parameter


justin-lau:
Currently the `minPitch` and `maxPitch` restrict the angle of the viewer regardless of the zoom level. i.e. if the `minPitch` and `maxPitch` are both set to `0`, the viewer center is locked vertically when the viewer zooms in.

mpetroff:
That's how the parameters are defined, so of course the view center is locked if `minPitch` and `maxPitch` are set to the same value. Implementing the behavior you suggest would be nice, but it's a lot more complicated. Both zoom level and viewer size need to be taken into account, and a behavior needs to be decided on for what happens when the viewer is resized. I though about adding this, but it's not currently high on my list of priorities.

justin-lau:
Hi @mpetroff, thanks for the quick reply. I'm very new to pannellum. I just think the current behaviour is unnatural to the users. It's good to know that you plan to work on it. I would like to contribute on it too when I can spare the time, but I haven't worked on computer graphics for 3 years, so it might take a while ;-)

jannecederberg:
@justin-lau do you want to test?

mpetroff:
I just changed the `minPitch` and `maxPitch` limits such that they apply to the edge of the viewer instead of the center and take zoom into account.

PR #169, opened 2016-06-19
Attempting to fix issue 92 (alter minPitch/maxPitch upon zooming)

jannecederberg:
This was my first attempt and I'm sure it could be improved but as of now it seems to work at least on my use-cases both with full 360x180 panoramas (for which it's not needed but importantly doesn't seem to break anything) and more importantly seems to work for panoramas that are not full 360x180 and where view is limited by `haov`, `vaov`, `minPitch` and `maxPitch` to avoid black areas from view.

**Sidenote regarding Git/GitHub and pull requests:**
Not being a Git-guru I didn't know how to get rid of the extraneous merge-commits; only commit `f6ee7cc` actually "does" anything to mpetroff/master.

How would I deal with a similar situation in the future where/if the exact pull request is not merged and my branch as a result "deviates" from upstream?

### DIFF
--- a/doc/json-config-parameters.md
+++ b/doc/json-config-parameters.md
@@ -72,6 +72,10 @@ If set to `false`, the zoom controls will not be displayed. Defaults to `true`.
 
 If set to `false`, zooming with keyboard will be disabled. Defaults to `true`.
 
+### `mouseZoom`
+
+If set to `false`, zooming with mouse wheel will be disabled. Defaults to `true`.
+
 
 ### `showFullscreenCtrl`
 
--- a/src/js/libpannellum.js
+++ b/src/js/libpannellum.js
@@ -464,7 +464,7 @@ function Renderer(container, image, imageType, dynamic) {
                     z = Math.cos(horizonRoll) * Math.cos(horizonPitch) * Math.sin(pitch) +
                         Math.cos(pitch) * (-Math.cos(yaw) * Math.sin(horizonPitch) +
                         Math.cos(horizonPitch) * Math.sin(horizonRoll) * Math.sin(yaw));
-                pitch = Math.asin(z);
+                pitch = Math.asin(Math.max(Math.min(z, 1), -1));
                 yaw = Math.atan2(y, x);
 
                 // Calculate roll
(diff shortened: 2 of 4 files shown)


jannecederberg:
Actually I'd rather cancel this pull request and redo it without the 4 unnecessary commit. I didn't want to delete and re-fork my branch though as I wasn't sure about possible side-effects. Any comments on how to accomplish the "branch-sync" ?

mpetroff:
Deleting and re-forking is definitely the easiest; there are no side effects that I know of. Otherwise, you'd have to change what's considered the `master` branch to the commit before your changes and make your changes a different branch, which isn't particularly straightforward.

mpetroff:
I've looked at the changes, and they don't really resolve the issue. For starters, the changes only affect zoom changes made with the scroll wheel, not the keyboard, zoom buttons, or API. I also don't think the `minPitch` and `maxPitch` values should be updated directly, since this causes rounding errors to accumulate. The viewer's aspect ratio should also be taken into account when doing scaling, and the limits shouldn't be tied to the initial `hfov` value.

jannecederberg:
True, I got too stoked about feeling it was working that I forgot about the other means of zooming.

So the most logical place for implementing the feature would be in the "static" `setHfov` function (or a function that's called from `setHfov`), or what do you think?

I don't feel like I fully understand what you're referring to with limits not being tied to the initial `hfov` value. Are you referring to this: 

> `var oldHfov = config.hfov;`
> `...`
> `var hfovDiff = (oldHfov - newHfov) / 2;`

mpetroff:
With regards to the limits being tied to the initial `hfov`, I mean that if the initial `hfov` is changed in the configuration, the limits are changed as well, since the limits are being scaled when the `hfov` changes. For example, say the initial `hfov` is 50 and the max pitch limit is also 50; if the initial `hfov` is changed to 100 and the user then zoom in such that the `hfov` is back at 50, the limit is now 75 instead of the initial 50.

I thought about issue #92 again and implemented it in f6e972a152af2686e6d3307c543b08fa26f52cf3 and a48c55565e4f3ee18da4c697d89a7782542435fd.

jannecederberg:
I pulled your changes; should the changes be backwards-compatible in regards to behaviour of configuration values? I mean, should panoramas that are not full 360x180 work the same way with the same `minPitch`, `maxPitch`, `haov` and `vaov` parameters and look the same and omit black areas the same way as before?

I'm asking because for me pulling the aforementioned changes actually drastically changes how non-full-panoramas look when using the same config values as before; the non-full-panoramas now appear very squished with lots of non-desired black areas showing.

jannecederberg:
I created these examples to illustrate what I mean:

- [My approach](https://dl.dropboxusercontent.com/u/13648061/pannellum-jc/issue-92/i.htm) (based on _jannecederberg/issue-92_)
- [Your approach](https://dl.dropboxusercontent.com/u/13648061/pannellum-mp/issue-92/i.htm) (based on _mpetroff/master_)

The configuration for the examples is the same across both example URL's, only the `src/js/pannellum.js` code changes. Before https://github.com/mpetroff/pannellum/commit/f6e972a152af2686e6d3307c543b08fa26f52cf3 and https://github.com/mpetroff/pannellum/commit/a48c55565e4f3ee18da4c697d89a7782542435fd these both worked the same way but now there's a big difference in perceived results.

My _jannecederberg/issue-92_ branch does not account for zoom from keyboard, API or UI buttons but for mousewheel it works mostly correctly to my eye.

I don't mean to be a smart-a** or anything like that, just trying to understand how to get it resolved.

jannecederberg:
Ok, thank you, I'll give it a try :)

jannecederberg:
Altered the params on the example and works great, thank you! Also the new approach is probably (or at least I feel) more intuitive to configure. Thank you again!

Issue #178, opened 2016-06-27
multires script not creating all files for zoom level 4


strarsis:
Input image is an usual 11280x5640 TIFF (8bit) file.
Latest multires script is used (master), default parameters used except --quality 100.
Script runs through without any errors, config.json is generated.

config.json is used with latest panellum viewer build (from master).
HTTP errors and artifacts due to missing tiles when zooming 
to zoom level 4 (max zoom for the input image).

A pattern in the failing request URLs is, that those tiles, which requests fail, 
are those of the upper (range to 7) row and/or column bound:
- d7_4.jpg
- b2_7.jpg
- r2_7.jpg
- u7_3.jpg
(... and so on)

Edit: After comparing the working output of a different, older multires output folder, 
I must conclude that it is an error of the multires script and not of the viewer!

Those missing files are the "seam" tiles, 90 in count.
When I add those tiles from an older working multires output folder, 
those seams are loaded + shown, so the multires script is skipping these seams.

```json
{
    "type": "multires",
    
    "multiRes": {
        "path": "/%l/%s%y_%x",
        "fallbackPath": "/fallback/%s",
        "extension": "jpg",
        "tileResolution": 512,
        "maxLevel": 4,
        "cubeResolution": 3584
    }
}
````

The multires script outputs correct tile files for zoom level 4 
with an input image of 11312x5656 - but not with 11280x5640.

Edit 2:
````
cubeSize = 8 * int(origWidth / math.pi / 8)
````
The working input image cubeSize is 3600, 
the troubled input image cubeSize is 3584.

````
tiles = int(math.ceil(float(size) / args.tileSize))
````
tiles for zoom level 4 is 8 for working input image and 7 for troubled input image.

Removing the rounding (int(...)) seems to fix the issue 
insofar that tiles is also 8 for the troubled input image:
````
cubeSize = 8 * (origWidth / math.pi / 8)
````

mpetroff:
There was an off-by-one error in `libpannellum.js` that was causing the issue.

Changing the cube size calculation in `generate.py` just avoided the bug for your input image by changing the cube size; it wasn't the cause of the problem.

strarsis:
Thanks! This bug caught me in surprise.

strarsis:
Just in case of a regression, I encountered this error and flickering, so in case someone ever wonders:
````
[...]

PR #179, opened 2016-06-27
Remove rounding in cubeSize calculation

strarsis:
Fixes https://github.com/mpetroff/pannellum/issues/178


mpetroff:
See comment in #178.

PR #189, opened 2016-07-03
fixed #188

### LINKED ISSUE #188: bug - config's "basePath" not used when loading via URL Configuration Parameters
Hi,

When using pannellum with url config file, the config "basePath" is ignored, that results on not found images.

Cheers

Cyrille37:


Cyrille37:
Argh. It does not do the job! Sorry.

Issue #196, opened 2016-07-15
Example of 'load' and 'scenechange' event listeners


rodinh:
Hi Matthew, 

First off, loving Pannellum viewer! I've experimented with various viewers, yours is the best I've come across.

As my users of the Pannellum tour are getting lost in the tour, they have suggested that I add a map/overview. In order to create a map I need the output of the `'load'` and '`scenechange`' event. 

I had no trouble in creating the tour, implementing the `'load' `and `'scenechange'` listeners is proving to be rather difficult for me. So sorry for asking you some 'newb'-ish questions. 

- I have instantiated the viewer
`viewer = pannellum.viewer('panorama', {my config is in here})`
- I have created a listener(?) via
`viewer.on('scenechange', sceneChangeListener)`
- JavaConsole then throws the following error: 
`"Uncaught ReferenceError: sceneChangeListener is not defined"`
-> So I understand I have to define sceneChangeListener somewhere? How would I properly set up the sceneChangeListener? 

Thanks in advance!

MmmDee:
@rodinh This works in my code:

```
viewer.on('scenechange',
	function () {
		console.log("Scene changed");
		console.log("Yaw " + viewer.getYaw());
		console.log("Pitch " + viewer.getPitch());
	}
);
```

MmmDee:
While on the topic of `scenechange` and `load` events... Matthew, it would be a nice enhancement to be able to have a bit more context available in those event handlers... such as viewer, scene type, hotSpot sceneId, etc.

rodinh:
@MmmDee Works like a charm! Thanks so much!

PR #234, opened 2016-08-26
Send the sceneId to 'scenechange' event

Aykelith:
In larger projects I need to know where the user clicked.


MmmDee:
Until/if the pull is approved, you can obtain the new sceneId using `getScene` within the `sceneChange` event. See issue #196 for an example.

mpetroff:
Thanks!

Issue #237, opened 2016-08-30
Dynamic adding hot spots and removing them


Aykelith:
Edit by @mpetroff: moved to #238.

PR #238, opened 2016-08-30
Dynamic adding hot spots and removing them

Aykelith:
I'm working on adding hot spots and removing them from the scene. The adding part is done and can be seen here: https://github.com/Aykelith/pannellum/commit/ae7a373c77070d64cad23d29198c88f51bda8cef where I create a function named `createHotSpot(hs)` (`hs` is the configuration for the hotspot) which create just an hotspot, actually is the exact same part of the `config.hotSpots` iteration of function `createHotSpots()` ([L1555-L1629](https://github.com/mpetroff/pannellum/blob/master/src/js/pannellum.js#L1555-L1629)) and the exact same thing I have done for `renderHotSpots()` by creating the function `renderHotSpot(hs)` ([L1661-L1687](https://github.com/mpetroff/pannellum/blob/master/src/js/pannellum.js#L1661-L1687)). And I created the public function `addHotSpot(hs)` which create just that hotspot, add it to the `config.hotspots` array and then only render it.

The problem comes when I need to create a function to delete a hotspot. The simplest way would be to just ask for the `index` of the hotpost, but it is not a nice way do to it for the user. A way would be to send, when creating the hotspot, an additional `id`( optional, of course) and in the function to delete the hotspot to ask for the hotspotId as a parameter and the function to return a boolean in case if he successfully found the hotpost and deleted it or not. This way the API will not be broken. I'm waiting for opinions.

EDIT: Added the remove function and pull requested the code. Feel free to rename the functions or the documentation.

Edit by @mpetroff: moved here from #237.

### DIFF
--- a/src/js/pannellum.js
+++ b/src/js/pannellum.js
@@ -2443,13 +2443,38 @@ this.getConfig = function() {
  * @memberof Viewer
  * @instance
  * @param {Object} hs - The configuration for the hotspot
- * @returns {Object} Configuration of current scene
+ * @returns {Viewer} `this`
  */
 this.addHotSpot = function(hs) {
     createHotSpot(hs);
     config.hotSpots.push(hs);
     renderHotSpot(hs);
     return this;
+}
+
+/**
+ * Remove the hotspot that has that id
+ * @memberof Viewer
+ * @instance
+ * @param {string} hotSpotId - The id of the hotspot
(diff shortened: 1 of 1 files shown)


mpetroff:
Thanks! I added this with some minor modifications in 7eae57bcc93515e05710e4d0bb1d9932e9fcfd12.

PR #241, opened 2016-08-31
Added functions to dynamic add/delete scenes

Aykelith:
The function to add scenes just add the scene to the `initialConfig.scenes` and the delete function is a boolean function that check if the scene to be deleted is not the current scene and if exists.


mpetroff:
Thanks again! Committed with minor modifications in 22e231a73119d9b8a48d5fb7d5e1a41a3b0e037c.

PR #242, opened 2016-09-01
Fixed the removeHotSpot function

Aykelith:
The function has an open `if` and the second argument of the `if` was missing.


Aykelith:
My inattention, sorry.

mpetroff:
It was confusing, since I accidentally inserted the comment in the wrong spot. I just fixed this.

PR #251, opened 2016-09-09
Added function to add hotspot to another scene

Aykelith:
I have added a function to add hotspot to another scene. The single problem is what to do if the scene is the current scene? A way would be to call `addHotSpot`( if can call it) or just return `false` and the user should figure out what he needs to do in that case.

EDIT: I forgot to add the check if `hotSpot` array does not exists.

### DIFF
--- a/src/js/pannellum.js
+++ b/src/js/pannellum.js
@@ -2455,6 +2455,23 @@ this.addHotSpot = function(hs) {
 }
 
 /**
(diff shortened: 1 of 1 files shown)


mpetroff:
Instead of merging this, I extended `addHotSpot` in b3123271124c5338554a408cb61ca7a52b153107 to have an optional `sceneId` parameter so it can be used to add hot spots to either the current scene or any other scene.

PR #302, opened 2016-11-22
Added doubleclick zoom functionality

vankessel:
I've ran some user tests on older users, and when asked to zoom in every single one of them tried to double click on the object of interest. I've added the doubleclick zoom-in as default functionality that can be turned off in the config. I've ensured that it does not interfere with rapidly clicking on UI elements.

I can change this functionality to default to off if you prefer. Also, my editor (Atom) deleted some whitespace automatically. I don't think it's a big deal but I can revert it and resubmit the pull request if needed.

Lastly, I've discovered a bug that has existed even before my changes. Double clicking on a scene change arrow will cause the panorama to not render properly. I will submit an issue for it.

### DIFF
--- a/src/js/pannellum.js
+++ b/src/js/pannellum.js
@@ -409,7 +409,7 @@ function onImageLoad() {
     if (!listenersAdded) {
         listenersAdded = true;
         dragFix.addEventListener('mousedown', onDocumentMouseDown, false);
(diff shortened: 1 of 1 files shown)


mpetroff:
Thanks! The whitespace changes were distracting, so I rebased this without them and added it in 37950068935e01a864d8c24f7540c995c9796c07.

PR #318, opened 2016-12-21
Provide fallback cursor

### LINKED ISSUE #317: Grab/grabbing cursors do not appear
The default pointer cursor is shown in some browsers while hovering/dragging instead of grabbing hands.

Specifically it appears Edge does not currently support using SVG data URIs.

A fallback cursor should be provided.

acortelyou:
Provide fallback grabbing hands for browsers without cursor svg data uri support.

Fixes #317.

### DIFF
--- a/src/css/pannellum.css
+++ b/src/css/pannellum.css
@@ -24,10 +24,12 @@
 }
 
 .pnlm-grab {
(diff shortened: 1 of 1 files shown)


mpetroff:
Thanks!

Issue #333, opened 2017-01-24
An option to enable mouse wheel zooming only in full screen mode.


stevenbg:
I'm using pannellum to show a panorama embeded in a web page and I set mouseZoom to false, allowing the user to scroll as expected. But when the user clicks the full screen button, the wheel no longer has another function and it is actually expected to zoom.

Sorry if I missed it, but I didn't find an onfullscreen event or a way to change configuration on the fly. Do you think pannellum should have those or a new config option for the described behaviour, which I believe is a common scenario?

mpetroff:
I think changing the available inputs methods depending on whether or not the viewer is fullscreen is quite confusing as far as usability goes, so I'm not sure this is a good idea. I tend to think the default behavior (with `mouseZoom` set to its default value of `true`) is the generally expected behavior, with the scroll wheel zooming the viewer when inside of it and scrolling the page when not over the viewer.

mpetroff:
That's not to say I won't add this, but I would need more convincing before doing so.

stevenbg:
I understand what seems a common scenario to me may not be so common and not worthy of a dedicated option. I suggested it, because while adding fullscreen events and a setConfig method is the better way, especially the latter seems like major work. Since nobody else in 3 years requested changing config on the fly, perhaps it's not worth it.

Currently I can do what I want messing directly with the DOM element of the viewer, which is dirty, but works.

mpetroff:
I thought about this some more. While I'm still not particularly fond of changing available input methods based on whether or not the viewer is fullscreen, I realized I could add the option by changing a single line of code. Starting with b0b7a77cbbfc033338fdd1b6f5c845106088e66d, the `mouseZoom` option can be set to `fullscreenonly` to enable mouse wheel zooming only while fullscreen.

As far as changing the configuration on the fly, this would be complicated to add because different parts of the configuration are processed at different times.

stevenbg:
Hello again!

The way `fullscreenonly` works at the moment is that when you are using the wheel to scroll the page and the cursor comes over the embedded panorama, it won't zoom the panorama, but it won't keep scrolling the page either because you are calling event.preventDefault().

Please see if you can accept my changes.

mpetroff:
Oops; that's what I get for testing with a page that isn't long enough to scroll.

[...]

PR #338, opened 2017-02-06
Allow scrolling when using the 'fullscreenonly' mode (part of #333)

stevenbg:
This change fixes the scrolling behavior for #333


mpetroff:
See https://github.com/mpetroff/pannellum/issues/333#issuecomment-278219063.

PR #339, opened 2017-02-08
Add draggable option

haripo:
Added an option to disable rotation by dragging and swiping.
I'm sorry if there is an existing way to do that :pray:


mpetroff:
Could you please explain why you'd like such an option? I'm not sure I see a use case for disabling the primary input method. I'd be happy to merge the pull request if I understood what it's for.

haripo:
I'm trying to create a navigation system, so I want to display a scenery of the right direction that the user should head for.
It's preferred that the user cannot rotate the scenery in my case.

mpetroff:
Makes sense. Thanks!

PR #345, opened 2017-02-19
Language configuration option for titles and hotspot text

campbellhennessy:
Hello,
This is my first pull request ever so my apologies if I've done anything wrong (please let me know and I'll try again).

I need multi-language support for titles and hotspot text for a project I am working on.  I've added an optional config entry 'language'.  If it is set to something (say 'fr'), pannellum will look for title_fr and text_fr for each scene and each hotspot.  If no language is set or no special title or text is included it will default to the title and text variables.

### DIFF
--- a/src/js/pannellum.js
+++ b/src/js/pannellum.js
@@ -1592,6 +1592,11 @@ function createHotSpot(hs) {
     var span = document.createElement('span');
     if (hs.text)
         span.innerHTML = escapeHTML(hs.text);
+        if(config.hasOwnProperty('language')){
+          if(hs.hasOwnProperty('text_' + config['language'])){
+            span.innerHTML = escapeHTML(hs['text_' + config['language']]);
+          }
+        }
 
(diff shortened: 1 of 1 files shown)


mpetroff:
There's nothing wrong with the pull request _per se_, but I disagree with how this it was implemented. The language setting needs to be changed in the configuration, so there needs to be multiple configurations with or without this feature. I think language specific configurations should be left to some sort of preprocessing step and not included in Pannellum directly, at least for the time being, be it separate configuration files on the server or a JavaScript function that prepares the configuration client-side.

At some point, I'd like to implement proper i18n support for Pannellum to allow everything to be translated. While this will be more work, I don't like the alternative of adding piecemeal translation support like is done in this pull request.

campbellhennessy:
Fair enough, thanks for taking a look :).  I'll probably eventually go with a pre-processing step as you suggest.  Great viewer btw.

PR #358, opened 2017-03-12
Add viewer reference in standalone

strarsis:
This PR adds a reference to the newly instantiated viewer so it can be controlled from the outside 
(and also from outside the iframe).


mpetroff:
Thanks! You missed the second place a viewer is created in the standalone viewer, though, so I added that in 6ba611cd75a1469ba1f48ea1dd0c18796b4b5b65.

PR #359, opened 2017-03-13
Make scrollable in touch only device when config.draggable is false

zcfan:
Meet a problem when using new draggable option.

I made a viewer take up entire viewport and rotate follow my cursor. (without click)

It works in desktop with mouse, but in mobile phone, I can't scroll down my page because viewer consumed the touchmove event.

Since viewer don't need to response to mousemove event when draggable is false, I think it's fine to simply return.


mpetroff:
Thanks!

zcfan:
No, thank you :-)  This cool project helps me a lot.

PR #410, opened 2017-06-12
Fix removing event listener inside listener

Tortila90:
It's a common situation to remove an event listener inside its definition. Like this:
```javascript
viewer.on('load', function foo() {
	//code
	viewer.off('load', foo);
});
```

In this case `off` function removes the item from `externalEventListeners` array, while looping through this array may be not finished in `fireEvent` function. A reverse iteration fixes it.
Since we'd like to fire listeners in the order they were added, we need to use `externalEventListeners[type].length - i` index.


mpetroff:
Thanks!

PR #414, opened 2017-06-21
Fix yaw out of bounds in mouseEventToCoords

Tortila90:
If config.yaw is close to -180 or 180, yaw might be out of bounds.


Tortila90:
Sorry, I'm new to GitHub. I don't know why my previous commit has been included as well. It had to be in previous request, and it was actually there. What did I do wrong?

mpetroff:
Thanks. I ended up manually applying the patches.

Issue #455, opened 2017-08-25
generate.py error


qwe5788319:
[error](http://chuantu.biz/t6/20/1503630779x993376187.png)

mpetroff:
Pillow used to automatically drop the alpha channel when saving a JPEG, but now it causes this error. Fixed in 972e7e4f3da421850ac870ef18a137489fa8091f.

PR #456, opened 2017-08-25
Multires: Convert images to RGB when necessary.

MikeN123:
Fixes mpetroff/panellum#455

Not sure if this works well for PNG images though, seems to work fine for JPEG.


mpetroff:
This works fine for PNG, but doesn't work for grayscale images. I fixed #455 in 972e7e4f3da421850ac870ef18a137489fa8091f.

PR #482, opened 2017-09-22
adding config to change loading text

fvcaputo:
This adds the option to change the text on the loading box when loading a panorama. It defaults to the old "Loading..." message, but can be changed to other texts. Helps if you want a different message, specially if you want it in another language. I figured `loadingLabel` is a good enough name for the config.

### DIFF
--- a/doc/json-config-parameters.md
+++ b/doc/json-config-parameters.md
@@ -170,6 +170,11 @@ Specifies the author to be displayed while the load button is displayed.
 Label to display on load button. Defaults to `Click to Load Panorama`.
 
 
+### `loadingLabel` (string)
+
+Label to display on the loading box. Defaults to `Loading...`.
+
+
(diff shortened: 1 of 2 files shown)


mpetroff:
Translation support is a needed improvement, but I don't want to take a piecemeal approach, such as this, by adding individual parameters for certain messages. There needs to be a way to translate all strings. I think the best way to do this is probably via a single dictionary parameter with an entry for each string; if a given entry is specified, it's used, else the default is used.

I appreciate the pull request, but since I don't want to take a piecemeal approach, I'm not going to merge it.

fvcaputo:
I see what you're saying. I also agree, maybe an approach with a dictionary parameter would be best, I'll see if I do something about it this week. Thank you.

mpetroff:
I just made all of the user-facing strings configurable in 93f3df1431fbd9b3eb5a6467b86b1f03e7afbf0d.

PR #556, opened 2018-02-26
Added config option to adjust the drag speed by a factor

jonathanrbowman:
We had someone asking if they could more fine-tune the speed at which the panorama drags when using touch screens in order to slow it down. This config option provides a way to scale it so you can increase/decrease the speed.

The config option is called `touchmovePanSpeedCoeffFactor`, and it defaults to 1 so as to not change existing functionality. Setting it to something lower, like .5, causes the panorama to drag more slowly, but you can also raise it if you want to translate smaller gestures into greater distance covered in panning.

Currently, I have it so this option only applies to touch and not mouse movement.


mpetroff:
Thanks! I renamed the parameter to `touchPanSpeedCoeffFactor` and documented it before merging.

jonathanrbowman:
@mpetroff Sure, sounds good and thanks for making such a great piece of software!

PR #563, opened 2018-03-03
support backgroundColor for Cube and Mulitres panoramas

DDvO:
This PR is going to extend the support for partial panoramas of any image type.

I tried extending the use of the configurable backgroundColor to image types other than equirectangular, but did not get far with that. Nevertheless I include first steps in this direction. Please note the two TODOs there.
Hopefully someone who is deeper into the geometry of cube and multires images will continue on this.

Nevertheless I will provide a PR that will restrict the viewing angles such that no parts of the full sphere with missing data (currently displayed as black-colored areas) are shown.

### DIFF
--- a/src/js/libpannellum.js
+++ b/src/js/libpannellum.js
@@ -317,6 +317,11 @@ function Renderer(container) {
         program.texCoordLocation = gl.getAttribLocation(program, 'a_texCoord');
         gl.enableVertexAttribArray(program.texCoordLocation);
 
+        // Set background color
+        program.backgroundColor = gl.getUniformLocation(program, 'u_backgroundColor');
+        var color = params.backgroundColor ? params.backgroundColor : [0, 0, 0];
+        gl.uniform4fv(program.backgroundColor, color.concat([1]));
+
         if (imageType != 'multires') {
             // Provide texture coordinates for rectangle
             if (!texCoordBuffer)
@@ -342,13 +347,6 @@ function Renderer(container) {
             gl.uniform1f(program.h, haov / (Math.PI * 2.0));
             gl.uniform1f(program.v, vaov / Math.PI);
             gl.uniform1f(program.vo, voffset / Math.PI * 2);
-
-            // Set background color
-            if (imageType == 'equirectangular') {
-                program.backgroundColor = gl.getUniformLocation(program, 'u_backgroundColor');
-                var color = params.backgroundColor ? params.backgroundColor : [0, 0, 0];
-                gl.uniform4fv(program.backgroundColor, color.concat([1]));
-            }
 
             // Create texture
             program.texture = gl.createTexture();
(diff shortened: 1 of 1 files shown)


mpetroff:
I merged this before I realized that it didn't actually add support for background colors for either cube or multires panoramas. As a patch, it provides a good starting point for someone who wants to work on this, but I don't think it serves much of a purpose in the main branch yet, so I reverted the change.

Figuring out what's in the current view is really tricky, which is why I haven't already added this.

DDvO:
I unterstand.
I just wanted to make sure that the effort spent in finding out how to extend the WebGL declarations does not get lost and someone else can build on it.
--
This message was sent using my Vernee Apollo Lite with K-9 Mail. Please excuse my brevity.

mpetroff:
After thinking about it some more, adding background color support for multires panoramas isn't difficult, just the fragment shader is the wrong place to do it. I added support in c979ff362ed8d39f05286bd22ddbb75cc8e47a9a by using the background color for the clear color. 

Supporting cube maps is trickier, since they currently don't work with missing images. If I modify `pannellum.js` to ignore the load error and edit `libpannellum.js` to only upload the textures for faces that have images, I get a WebGL error, so I think WebGL requires an image texture for each face of a cube map texture. Thus, I think adding support would probably require generating an image of the specific color and uploading it as the face texture. I'm not sure it's worth it, since one could just create a solid colored PNG (which is <1KB) and use it for the missing faces.

DDvO:
This would be very nice, but does not (yet) work for me.

I just tried out the new version: on my partial multires test pano I still get the black background 
(although I set the option `backgroundColor: [0.5, 0.5, 0.5]` and i verified that the option was indeed used in line 325).

mpetroff:
That's odd. Here's what I get when I set the background color for the multires example and delete the `f00` tile:

screen shot 2018-03-04 at 14 17 08

It works fine for me on both Firefox and Chromium.

DDvO:
A striking example! 
Meanwhile I understand why it does not work for me as expected:
The tiles of [my pano](http://gallery.davoh.de/Seebensee/dsc_3927-3933.html), as generated by nona (in my extended `generate.gy`) of course still have those black portions past the edges of the original image, and these are shown. When I make min/maxPitch and min/maxYaw large enough, I do get shown the newly set background color for missing tile images.

So the current solution is limited to the (non-)existence of tile files. 
It should better be based on the geometry defined by min/maxPitch and min/maxYaw.

mpetroff:
Defining it by geometry is complicated. Unless you really need to be able to dynamically change the background color, it's much simpler to set the background color when generating the tiles and using the background color parameter to set the color where there aren't tiles.

DDvO:
I see. I was already looking for an option of `nona` like `backgroundColor`, but could not find one.
Can you tell me how to achieve that?

DDvO:
https://github.com/mpetroff/pannellum/pull/567 contains a different approach for partial panoramas:

prevent display of out-of-range areas (i.e, outside [minYaw, maxYaw] or [minPitch, maxPitch]) of partial panoramas by adapting yaw and horizontal fov, thus the use of backgroundColor is confined to any missing tiles within the image boundaries.

mpetroff:
When using an output format that supports transparency, e.g. TIFF as used by `generate.py`, `nona` makes the background transparent. When removing the transparency with Pillow, the transparent areas default to black, but this can be changed; see https://github.com/python-pillow/Pillow/issues/2609#issuecomment-313922483 (Pillow's [color string reference](http://pillow.readthedocs.io/en/3.1.x/reference/ImageColor.html#color-names) might also be helpful).

DDvO:
Thanks @mpetroff for the good hint regarding setting the background color in images generated with `nona`. I've been able to extend the `generate.gy` script accordingly and will contribute this (together with further extensions of the script supporting partial panos and cylindrical) input ASAP.

PR #564, opened 2018-03-03
fix a copy&paste error on check for movement stop if opposite controls are pressed

DDvO:
This is a trivial fix for an apparently minor bug w.r.t. checking elements of `keysDown`.

It also highlights that two functions are unused: `getYawBounds()` and `getYawBounds()`.


mpetroff:
Good catch on the typo! Thanks.

The two functions you marked are API functions, none of which are used internally, so I don't see why they should be marked (I removed this part of the commit).

DDvO:
I see, fine.
--
This message was sent using my Vernee Apollo Lite with K-9 Mail. Please excuse my brevity.

PR #567, opened 2018-03-04
avoid showing out-of-range areas of partial (cylindrical) panoramas

DDvO:
By adapting yaw and horizontal FOV,
out-of-range areas (i.e, outside [minYaw, maxYaw] or [minPitch, maxPitch]) of partial panoramas are never shown.

### DIFF
--- a/src/js/pannellum.js
+++ b/src/js/pannellum.js
@@ -99,6 +99,7 @@ var defaultConfig = {
     orientationOnByDefault: false,
     hotSpotDebug: false,
     backgroundColor: [0, 0, 0],
+    avoidShowingBackground: false,
     animationTimingFunction: timingFunction,
     draggable: true,
     disableKeyboardCtrl: false,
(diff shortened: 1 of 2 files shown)


src/js/pannellum.js
```diff
@@ -2117,13 +2136,23 @@ function constrainHfov(hfov) {
...
         return config.hfov;
-    } if (hfov < minHfov) {
-        return minHfov;
+    }
+    var newHfov = config.hfov;
+    if (hfov < minHfov) {
+        newHfov = minHfov;
     } else if (hfov > config.maxHfov) {
-        return config.maxHfov;
+        nwHfov = config.maxHfov;
```
mpetroff: I assume this is a typo.

src/js/pannellum.js
```diff
@@ -1400,23 +1401,41 @@ function render() {
         // Keep a tmp value of yaw for autoRotate comparison later
         tmpyaw = config.yaw;
 
+        // prevent showing background (empty space) on left or right by adapting min/max yaw
+        var canvas = renderer.getCanvas(),
+            canvasWidth = canvas.width / (window.devicePixelRatio || 1),
+            canvasHeight = canvas.height / (window.devicePixelRatio || 1),
```
mpetroff: Since you're only using the width and height to calculate the aspect ratio, there's no need for using the device pixel ratio.

src/js/pannellum.js
```diff
@@ -2117,13 +2136,23 @@ function constrainHfov(hfov) {
...
+        nwHfov = config.maxHfov;
     } else {
-        return hfov;
+        newHfov = hfov;
+    }
+    if (renderer) {
+        // prevent showing background (empty space) on top or bottom by adapting newHfov
+        newHfov = Math.min(newHfov,
+                           Math.atan(Math.tan((config.maxPitch - config.minPitch) / 360 * Math.PI) /
+                                     renderer.getCanvas().height * renderer.getCanvas().width)
```
mpetroff: It would be better to assign the result of `getCanvas()` to a variable instead of calling it twice.

src/js/pannellum.js
DDvO: Good point

src/js/pannellum.js
DDvO: oops

src/js/pannellum.js
DDvO: Indeed

DDvO:
Thanks for the hints; done.

mpetroff:
Thanks!

PR #570, opened 2018-03-11
support also partial cubemap/fallback images: backgroundColor for missing face(s)

DDvO:
Similarly to the recently added support for missing tiles in multires panoramas,
this PR adds support for missing faces in cubemap (which includes fallback) panoramas. 
Any areas for which no data is available are shown with the configurable `backgroundColor`.

I updated the documentation of `backgroundColor` accordingly.

This PR also includes a couple of minor improvements of related code.

This is in a sense a continuation of PR #563.

### DIFF
--- a/src/js/libpannellum.js
+++ b/src/js/libpannellum.js
@@ -102,11 +102,11 @@ function Renderer(container) {
 
         var s;
         var faceMissing = false;
-        var cubeImgWidth = undefined;
+        var cubeImgWidth;
         if (imageType == 'cubemap') {
             for (s = 0; s < 6; s++) {
                 if (image[s].width > 0) {
-                    if (cubeImgWidth == undefined)
+                    if (cubeImgWidth === undefined)
                         cubeImgWidth = image[s].width;
                     if (cubeImgWidth != image[s].width)
                         console.log('Cube faces have inconsistent widths: ' + cubeImgWidth + ' vs. ' + image[s].width);
@@ -242,7 +242,7 @@ function Renderer(container) {
             };
             var incLoaded = function() {
                 if (this.width > 0) {
-                    if (fallbackImgSize == undefined)
+                    if (fallbackImgSize === undefined)
                         fallbackImgSize = this.width;
                     if (fallbackImgSize != this.width)
                         console.log('Fallback faces have inconsistent widths: ' + fallbackImgSize + ' vs. ' + this.width);
--- a/src/js/libpannellum.js
+++ b/src/js/libpannellum.js
@@ -101,6 +101,37 @@ function Renderer(container) {
         pose = undefined;
 
         var s;
+        var faceMissing = false;
+        var cubeImgWidth = undefined;
+        if (imageType == 'cubemap') {
+            for (s = 0; s < 6; s++) {
+                if (image[s].width > 0) {
+                    if (cubeImgWidth == undefined)
+                        cubeImgWidth = image[s].width;
+                    if (cubeImgWidth != image[s].width)
+                        console.log('Cube faces have inconsistent widths: ' + cubeImgWidth + ' vs. ' + image[s].width);
+                } else
+                    faceMissing = true;
+            }
+        }
+        function fillMissingFaces(imgSize) {
+            if (faceMissing) { // fill any missing fallback/cubemap faces with background
+                var nbytes = imgSize * imgSize * 4; /* RGB, plus non-functional alpha */
+                var imageArray = new Uint8ClampedArray(nbytes);
+                var [r,g,b] = params.backgroundColor ? params.backgroundColor : [0, 0, 0];
+                r *= 255; g *= 255; b *= 255;
+                var i;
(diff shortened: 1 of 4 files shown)


DDvO:
I've continued this branch with a major extension of `utils/multires/generate.py` to cylindrical input and partial panoramas.

The tile and fallback face files will only be produced when they actually contain image data.
For those files produced, the background color used past the edges of a partial pano can be defined.

I also added some generally useful options. Some options are just passed through in the .json file to the viewer.

Added options regarding partial panoramas:
- -b --backgroundcolor: RGB triple of values [0, 1] defining background color shown past the edges of a partial panorama (defaults to "[0.0, 0.0, 0.0]")
- -B --avoidbackground: viewer should limit view to avoid showing background, so using --backgroundcolor is not needed
- -H --haov: horizontal angle of view (defaults to 360.0 for full panorama)
- -V --vaov: vertical angle of view (defaults to 180.0 for full panorama) 

Added options regarding cylindrical input:
- -C --cylindrical: input projection is cylindrical (default is equirectangular)
- -e --horizon: offset of the horizon in pixels (negative if above middle, defaults to 0)

Added general options:
- -F --hfov: starting horizontal field of view (defaults to 100.0)
- -O --voffset: starting pitch position (defaults to 0.0)
- -f --fallbacksize: fallback tile size in pixels (defaults to 1024)
- -a --autoload: automatically load panorama in viewer
- -G --gpu: perform image remapping by nona on the GPU
- -d --debug: debug mode (print status info and keep intermediate files

src/js/pannellum.js
```diff
@@ -350,6 +350,11 @@ function init() {
         };
         
         var onError = function(e) {
+            if (1) { // support partial cubemap image, i.e., missing faces
+                console.log(config.strings.fileAccessError.replace('%s', e.target.src)+'; will use background instead');
+                onLoad();
+                return;
+            }
```
mpetroff: This prevents helpful error messages when URLs are incorrect. Missing faces should be defined as `null` in the configuration, and this should then be checked instead of trying and failing to load the face.

utils/multires/generate.py
```diff
@@ -42,59 +43,107 @@
...
-levels = int(math.ceil(math.log(float(cubeSize) / args.tileSize, 2))) + 1
+tileSize = min(args.tileSize, cubeSize)
+levels = int(math.ceil(math.log(float(cubeSize) / tileSize, 2))) + 1
 origHeight = str(origHeight)
 origWidth = str(origWidth)
 origFilename = os.path.join(os.getcwd(), args.inputFile)
 extension = '.jpg'
 if args.png:
     extension = '.png'
+colorList = eval(args.backgroundColor)
```
mpetroff: Calling `eval` is unsafe; `ast.literal_eval` should be used instead.

utils/multires/generate.py
```diff
@@ -42,59 +43,107 @@
...
+
+if args.debug:
+    print('maxLevel: '+ str(levels))
+    print('tileResolution: '+ str(tileSize))
+    print('cubeResolution: '+ str(cubeSize))
 
 # Create output directory
-os.makedirs(args.output)
+if not os.path.exists(args.output):
+    os.makedirs(args.output)
```
mpetroff: This was intended to fail if the output directory already exists. A check could be added to print a descriptive error message instead of a stack trace, but I don't think the script should continue if the output directory already exists.

src/js/libpannellum.js
```diff
@@ -101,6 +101,37 @@ function Renderer(container) {
         pose = undefined;
 
         var s;
+        var faceMissing = false;
+        var cubeImgWidth = undefined;
```
mpetroff: The `= undefined` part is unnecessary; `var cubeImgWidth;` is functionally identical.

src/js/libpannellum.js
```diff
@@ -101,6 +101,37 @@ function Renderer(container) {
...
+        if (imageType == 'cubemap') {
+            for (s = 0; s < 6; s++) {
+                if (image[s].width > 0) {
+                    if (cubeImgWidth == undefined)
```
mpetroff: Strict comparison, `===`, should be used.

src/js/libpannellum.js
```diff
@@ -206,38 +238,44 @@ function Renderer(container) {
...
-                incLoaded();
+                incLoaded.call(this);
             };
             var incLoaded = function() {
-                if (this.width != 0) // support partial fallback/cubemap image
-                    fallbackImgSize = this.width;
-                if (loaded == 5 && this.width == 0) // support partial fallback/cubemap image
-                    this.width = fallbackImgSize;
+                if (this.width > 0) {
+                    if (fallbackImgSize == undefined)
```
mpetroff: Strict comparison, `===`, should be used.

DDvO:
Thanks for your swift feedback; 
I've just done all requested changes.

mpetroff:
Great. Thanks!

Issue #175, opened 2016-06-23
Run from local file system if image data is supplied directly in the js code or in the HTML


spaceChRiS:
Pannellum looks like a great piece of software and it is almost what I search for. The only restriction that prevents my use case is that it cannot be run from local file system since js has no access to local files outside the DOM scope. I wonder if it is possible to add the image data directly into the js code or the DOM (e.g. as Data-URI in the HTML) to make it run locally, From local file system, neither the load speed nor the size of the image data is an issue, so that multiscale etc. would not be needed.

hadefix:
I would also be interessted if it is possible to stuff in the file data as data-url... the following seems not to work: 
"panorama": "data:image/JPEG;base64,FileData_64Encoded",

holzdev:
I have successfully used the apache cordova wrapper around the html5 file api to load files from local file system. Maybe using the file api could work in your use case.

spaceChRiS:
Thanks for the suggestion. As far as I understand, cordova is a tool to write platform-independent (almost, Linux seems not to be a target platform) applications based on web technology. How will it help me? Can you give some advice, please (disclaimer: I am not a programmer, but maybe able to write some code if necessary, so please be detailed & patient ;-) ).

My use case: I want to distribute a tour of panoramas as simple directory, e.g. index.html + supporting files. The receiver should be able to load the index.html and view the panoramas. No app, and totally platform independent, as long as the browser is recent enough, and perfect for archival purposes.

holzdev:
The focus is in the file api. Not Cordova. The file api is a means for the browser to access the local file system. You may be able to use it in order to load the images. Though this will require you to alter pannellum since it tries to load files via XMLHttpRequest which is where the CORS issues arise. 
I am not 100% sure it can be done but I don't see why not. Maybe the maintainer knows otherwise...


-------- Ursprüngliche Nachricht --------
Von: spaceChRiS <notifications@github.com> 
Datum: 19.07.2016  12:19 AM  (GMT+01:00) 
An: mpetroff/pannellum <pannellum@noreply.github.com> 
Cc: holzdev <paul.holz@posteo.de>, Comment <comment@noreply.github.com> 
Betreff: Re: [mpetroff/pannellum] Run from local file system if image data is
  supplied directly in the js code or in the HTML (#175) 

[...]

PR #573, opened 2018-03-11
enable viewing locally defined panoramas like the examples

### LINKED ISSUE #572: Enable the examples to run directly from local file system
As mentioned in #175, in many cases it is possible render panoramas from local files by using relative paths with no 'upward pointing' components (`../`) the browser usually allows access to local files. 

Unfortunately, the HTML files used for the Pannellum examples, e.g., `examples/example.htm`, as well as the support file `src/standalone/pannellum.htm`, use such upward references to JavaScript (`.js`), CSS, JSON, and HTML files.

I suggest to simply move `src/standalone/pannellum.htm` as well as the `.htm` files in the `example/` folder to the root folder of the repository, adapting the file references contained in them, and to disable/remove the check `window.location.protocol == 'file:'` in `src/standalone/standalone.js`. 

Then on many systems the examples can be conveniently run locally, with the need for workarounds like setting up a local HTTP server.

DDvO:
Fixes #572 (and likely also #175)


mpetroff:
I think this is likely to cause more confusion, since it only works in certain cases. The current error message at least makes it clear why it doesn't work. I also don't want to move `pannellum.htm` to the root directory.

I'm going to have to think about this more.

mpetroff:
Have you tested this in Chrome? I can only get it to work in Firefox.

kanalasumant:
Awesome, you guys seem to be on top of this. How about opening an issue with the chromium, so that in the future they may incorporate this feature without any feature flags. If they do not wanna proceed with this proposal, that's their call, but asking when needed seems to be the next logical thing here. @mpetroff @DDvO what do you think about this.

DDvO:
At least at the moment I do not have the time to discuss with Chrome/Chromium developers the underlying security implications, which appear rather intricate.

I've just improved the workaround for Pannellum proposed in this pull request.
Now the error message is given only if local file access actually has been denied.

I tested also with other browsers: Safari acts like Firefox (i.e., it does not care),
while Opera behaves like Chrome and provides the same command-line switch.

PR #600, opened 2018-04-27
Support parallel requests for multiRes panorama tiles

thenickdude:
Looking at the network waterfall diagram in Chrome's developer tab, it's clear that the individual tiles for multiRes panoramas are always loaded serially, making them excruciatingly slow to load if the latency to the server is high:

screen shot 2018-04-27 at 9 15 49 pm

This is true regardless of the setting of `cacheTop` in `loadTexture()`, which would ideally be in control of the texture loading parallelism:

```js
var cacheTop = 4;   // Maximum number of concurrents loads
```

The reason is due to this code:

```js
program.currentNodes.sort(multiresNodeRenderSort);
// Only process one tile per frame to improve responsiveness
for (i = 0; i < program.currentNodes.length; i++) {
    if (!program.currentNodes[i].texture) {
        console.log("setTimeout: " + program.currentNodes[i].path); // <- Added this for debugging
        setTimeout(processNextTile, 0, program.currentNodes[i]);
        break;
    }
}
```

Every frame, a list of nodes in the viewport is sorted by their position and level, and then because of the `break` statement, only the first of the incomplete nodes will ever be queued for loading. The sorting means that the same node is likely to be chosen for loading every time, unless the view shifts significantly, even if that node is already queued for loading. A new node is only chosen once the first node is loaded (`texture` gets set).

The net result is that the same tile gets queued over and over again (20 times or more). The cache loader pool is not properly utilised (only one or two loaders is ever run in parallel).

This patch allows nodes to be queued up for loading until the texture loading pool is completely full. After the patch, the network diagram on Chrome clearly shows 4 requests in the (pending) state simultaneously, and the panorama loading is astonishingly faster on high latency connections (100ms+).

screen shot 2018-04-28 at 9 39 43 am

### DIFF
--- a/src/js/libpannellum.js
+++ b/src/js/libpannellum.js
@@ -660,12 +660,29 @@ function Renderer(container) {
                 var ntmp = new MultiresNode(vtmps[s], sides[s], 1, 0, 0, image.fullpath);
                 testMultiresNode(rotPersp, ntmp, pitch, yaw, hfov);
             }
+            
             program.currentNodes.sort(multiresNodeRenderSort);
-            // Only process one tile per frame to improve responsiveness
-            for (i = 0; i < program.currentNodes.length; i++) {
-                if (!program.currentNodes[i].texture) {
-                    setTimeout(processNextTile, 0, program.currentNodes[i]);
-                    break;
+            
+            // Unqueue any pending requests for nodes that are no longer visible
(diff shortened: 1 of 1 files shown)


mpetroff:
Thanks! The multires tile loading is definitely something that's suboptimal.

PR #613, opened 2018-06-03
Removed wrong encodings of URLs (fixes #611).

### LINKED ISSUE #611: Wrong URI component decoding
Hi,

I'm trying to use Pannelum with images stored on Firebase Storage, but I'm having weird url decoding from both the JS api and the iframe component.

I've made a repository reproducing the errors described below: https://github.com/gcangussu/pannelum-decoding-bug

You can serve the files of this repository locally to see the errors.

There is also a firebase project with a file stored there for testing:
```
https://firebasestorage.googleapis.com/v0/b/pannellum-escaping-problem.appspot.com/o/Path%2FTo%2FFile%2Falma.jpg?alt=media&token=26371c15-87e8-4dc2-842b-fea688e817ec
```

I can't use this URL without any type of escaping with the pannellum iframe, because if I do this
```html
<iframe width="600" height="400" allowfullscreen style="border-style:none;" src="https://cdn.pannellum.org/2.4/pannellum.htm?panorama=https://firebasestorage.googleapis.com/v0/b/pannellum-escaping-problem.appspot.com/o/Path%2FTo%2FFile%2Falma.jpg?alt=media&token=26371c15-87e8-4dc2-842b-fea688e817ec"></iframe>
```
The `&token=26371[...]` will be considered part of the main url `https://cdn.pannellum[...]`. (See [iframe.html](https://github.com/gcangussu/pannelum-decoding-bug/blob/master/iframe.html))

So I tried using the JS api:
```js
pannellum.viewer('panorama', {
    "type": "equirectangular",
    "panorama": "https://firebasestorage.googleapis.com/v0/b/pannellum-escaping-problem.appspot.com/o/Path%2FTo%2FFile%2Falma.jpg?alt=media&token=26371c15-87e8-4dc2-842b-fea688e817ec"
});
```
~~But then I get this error: (see [js-api.html](https://github.com/gcangussu/pannelum-decoding-bug/blob/master/js-api.html))
js-api-pannellum~~

~~Note that Pannellum is trying to get the wrong URL, the `%252F` parts should be `%2F`. It encoded `%` to `%25`, but it should not do that, it breaks the URL.~~ (see update on following comment.)

I also tried using the iframe, but encoding the image url with `encodeURIComponent` (see [iframe-escaped.html](https://github.com/gcangussu/pannelum-decoding-bug/blob/master/iframe-escaped.html)). But then I have other problem:
iframe-escaped

Now Pannellum is decoding `%252F` to `/`. But it should not do that, it should only decode one time, from `%252F` to `%2F`. You can check the encoded url runnig this on your console
```
>> encodeURIComponent('https://firebasestorage.googleapis.com/v0/b/pannellum-escaping-problem.appspot.com/o/Path%2FTo%2FFile%2Falma.jpg?alt=media&token=26371c15-87e8-4dc2-842b-fea688e817ec');
"https%3A%2F%2Ffirebasestorage.googleapis.com%2Fv0%2Fb%2Fpannellum-escaping-problem.appspot.com%2Fo%2FPath%252FTo%252FFile%252Falma.jpg%3Falt%3Dmedia%26token%3D26371c15-87e8-4dc2-842b-fea688e817ec"
```
It transforms `%2F` to `%252F`, and if we decode it with `decodeURIComponent` we get our URL back:
```
>> decodeURIComponent("https%3A%2F%2Ffirebasestorage.googleapis.com%2Fv0%2Fb%2Fpannellum-escaping-problem.appspot.com%2Fo%2FPath%252FTo%252FFile%252Falma.jpg%3Falt%3Dmedia%26token%3D26371c15-87e8-4dc2-842b-fea688e817ec")
"https://firebasestorage.googleapis.com/v0/b/pannellum-escaping-problem.appspot.com/o/Path%2FTo%2FFile%2Falma.jpg?alt=media&token=26371c15-87e8-4dc2-842b-fea688e817ec"
```

I didn't look yet at Pannellum's implementation and of how it handles URL parameters, but I think it should use the standard compliant `decodeURIComponent` with the iframe component.

To conclude, I think these behaviours ~~with the JS api~~ and with the iframe+`encodeURIComponent` are bugs. I can help fixing them if they are confirmed.

Thank you,
Gabriel

gcangussu:
Fixes #611 

- Fixed double encoding with `encodeURI`.
- Added protection against `javascript:` URLs, which could be used for XSS.
- Changed use of `innerHTML` to `textContent` where it could be exploited for XSS.
- Removed double decoding of URL parameters in standalone module. I think the right approach is to decode each URL parameter isolated.
- Removed unnecessary array by using `location.search`.

### DIFF
--- a/src/js/pannellum.js
+++ b/src/js/pannellum.js
@@ -392,7 +392,7 @@ function init() {
                 if (xhr.status != 200) {
                     // Display error if image can't be loaded
                     var a = document.createElement('a');
-                    a.href = encodeURI(p);
+                    a.href = p;
                     a.innerHTML = a.href;
                     anError(config.strings.fileAccessError.replace('%s', a.outerHTML));
                 }
--- a/src/js/pannellum.js
+++ b/src/js/pannellum.js
@@ -352,7 +352,7 @@ function init() {
         var onError = function(e) {
             var a = document.createElement('a');
             a.href = e.target.src;
-            a.innerHTML = a.href;
+            a.textContent = a.href;
             anError(config.strings.fileAccessError.replace('%s', a.outerHTML));
         };
         
@@ -367,7 +367,7 @@ function init() {
                 }
                 panoImage[i].onload = onLoad;
                 panoImage[i].onerror = onError;
-                panoImage[i].src = encodeURI(p);
+                panoImage[i].src = sanitizeURL(p);
             }
         }
     } else if (config.type == 'multires') {
@@ -393,7 +393,7 @@ function init() {
                     // Display error if image can't be loaded
                     var a = document.createElement('a');
                     a.href = p;
-                    a.innerHTML = a.href;
+                    a.textContent = a.href;
                     anError(config.strings.fileAccessError.replace('%s', a.outerHTML));
                 }
                 var img = this.response;
@@ -1684,7 +1684,7 @@ function createHotSpot(hs) {
             p = hs.video;
         if (config.basePath && !absoluteURL(p))
             p = config.basePath + p;
-        video.src = encodeURI(p);
+        video.src = sanitizeURL(p);
         video.controls = true;
         video.style.width = hs.width + 'px';
         renderContainer.appendChild(div);
@@ -1694,11 +1694,11 @@ function createHotSpot(hs) {
         if (config.basePath && !absoluteURL(p))
             p = config.basePath + p;
         a = document.createElement('a');
-        a.href = encodeURI(hs.URL ? hs.URL : p);
+        a.href = sanitizeURL(hs.URL ? hs.URL : p);
         a.target = '_blank';
         span.appendChild(a);
         var image = document.createElement('img');
-        image.src = encodeURI(p);
+        image.src = sanitizeURL(p);
         image.style.width = hs.width + 'px';
         image.style.paddingTop = '5px';
         renderContainer.appendChild(div);
@@ -1706,7 +1706,7 @@ function createHotSpot(hs) {
         span.style.maxWidth = 'initial';
     } else if (hs.URL) {
         a = document.createElement('a');
-        a.href = encodeURI(hs.URL);
+        a.href = sanitizeURL(hs.URL);
         a.target = '_blank';
         renderContainer.appendChild(a);
         div.className += ' pnlm-pointer';
@@ -1930,7 +1930,7 @@ function processOptions(isPreview) {
             p = config.basePath + p;
         preview = document.createElement('div');
         preview.className = 'pnlm-preview-img';
-        preview.style.backgroundImage = "url('" + encodeURI(p) + "')";
+        preview.style.backgroundImage = "url('" + sanitizeURLForCss(p) + "')";
         renderContainer.appendChild(preview);
     }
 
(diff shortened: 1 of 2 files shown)


src/standalone/standalone.js
```diff
@@ -10,17 +10,16 @@ function parseURLParameters() {
     var URL;
     if (window.location.hash.length > 0) {
         // Prefered method since parameters aren't sent to server
-        URL = [window.location.hash.slice(1)];
+        URL = window.location.hash.slice(1);
     } else {
-        URL = decodeURI(window.location.href).split('?');
```
gcangussu: If we decode the whole URI before extracting the parameters we risk looping through parameters that are from query string variables such as `panorama`. And we will also not get the right URL for `panorama` or other parameter that is a URL.

src/js/pannellum.js
```diff
@@ -392,7 +392,7 @@ function init() {
                 if (xhr.status != 200) {
                     // Display error if image can't be loaded
                     var a = document.createElement('a');
-                    a.href = encodeURI(p);
```
gcangussu: This breaks the link and misleads the reason of the error to a wrong URL request.

mpetroff: Can you guarantee that this doesn't introduce the potential for DOM XSS attacks? That's why the encoding was added in the first place. My cursory attempt to demonstrate an attack failed, but I'd like to be more confident that this doesn't introduce an attack vector.

gcangussu: Good thinking. We need to prevent urls starting with `javascript:` protocol and use `.textContent` instead of `.innerHTML`.

But other types of XSS attacks are a problem only when you're using a template engine on the server. So there is no chance that the browser will get a `<a href="/" onclick=alert('xss')>` when doing `a.href = x`.

Latter we're setting `a.innerHTML` to `a.href`, `a.href` could have HTML tags, so I will change that to `a.textContent = a.href`. Using textContent prevents XSS attacks that could happen with innerHtml ([see rule 6](https://www.owasp.org/index.php/DOM_based_XSS_Prevention_Cheat_Sheet))

gcangussu: So I took the liberty to fix other places that could suffer from double encoding URLs and also replaced `encodeURI` with a simple sanitizer to prevent `javascript:` URLs.

src/js/pannellum.js
```diff
@@ -1971,7 +1971,16 @@ function processOptions(isPreview) {
                 break;
             
             case 'fallback':
-                infoDisplay.errorMsg.innerHTML = '<p>Your browser does not support WebGL.<br><a href="' + encodeURI(config[key]) + '" target="_blank">Click here to view this panorama in an alternative viewer.</a></p>';
```
gcangussu: This could be a source of XSS without `encodeURI`. So I changed to get the browser protections.

src/standalone/standalone.js
```diff
@@ -1,7 +1,9 @@
 function anError(error) {
     var errorMsg = document.createElement('div');
     errorMsg.className = 'pnlm-info-box';
-    errorMsg.innerHTML = '<p>' + error + '</p>';
```
gcangussu: I also modified here to avoid XSS.

mpetroff:
The XSS protection seems robust now. Thanks!

kaYcee:
We encountered an issue with our Google Cloud Storage links. Google signs image urls and appends the signature to the download link to ensure authenticity. This results in a mixture of URI-encoded and -decoded parts, e.g. `https://storage.googleapis.com/abcde/images%2Fpanoramas%2Fmypicture.jpg?GoogleAccessId=abcde%40appspot.gserviceaccount.com&Expires=16700323200&Signature=AI8iu9zev1xz6Qt%2BeqXk4%3D%3D...`. Up to now, this hasn't been a problem. But since Pannellum uses `encodeURI()` everywhere, either the front part of the url or its signature gets destroyed. There is no way we can feed it with a proper url, meaning `decodeURI()` prior to the call is not of much help here. 

The only workaround we came up with so far was replacing all occurances of `encodeURI()` with a non-altering version. Is there a way to make the encoding optional like the `escapeHTML` parameter value?

gcangussu:
@kaYcee I had a very similar problem, you should be good once this gets released.

Issue #284, opened 2016-11-14
Issue on mobile?


PTthe13:
Hey,

I'm having an issue implementing the **Pannellum** on **Ionic framework**.

I'm calling the pannellum information on a certain view _(the idea is for each view to have a different 360 image)_.

My problem is whenever I open the view I want, it returns the following error in the console: **_"Uncaught TypeError: g.addEventListener is not a function"_** and it never shows the pic (even though It apparently loads it).

![screen shot 2016-11-14 at 12 21 46](https://cloud.githubusercontent.com/assets/3495263/20264749/13fcf4bc-aa65-11e6-97ee-57ff3e4c5b84.png)
![screen shot 2016-11-14 at 12 21 40](https://cloud.githubusercontent.com/assets/3495263/20264751/14010994-aa65-11e6-84ca-7bc7b36b395e.png)
![screen shot 2016-11-14 at 12 22 08](https://cloud.githubusercontent.com/assets/3495263/20264750/13fdc00e-aa65-11e6-96fc-e5001b293529.png)

**_PS: I forgot to mention. While developing locally and testing on the browser, everything works fine. Only when I "push" to the devices (android and iOS) does it return that error._**

mpetroff:
Are you trying to run it on some sort of ancient version of WebKit? The error says that the [FileReader](https://developer.mozilla.org/en-US/docs/Web/API/FileReader) object doesn't have an `addEventListener` method. This was implemented in WebKit [almost five years ago](https://bugs.webkit.org/show_bug.cgi?id=42723).

If you're using an actual modern browser, my guess is that the issue is being caused by an interaction with some other piece of code, such as a framework.

PTthe13:
@mpetroff It's running on the most modern version of WebKit. 

It must be something to do with the Ionic framework, cus after I wrote that, I was testing around, and found out that if I "forced" the view to load no dynamic content and had the pannellum hardcoded, it worked fine.

I also found out that using the iframe mode instead of the API works fine.

It must be some kind of conflict with the Ionic framework. I'll be testing some more on this.

mpetroff:
It seems Ionic [reimplements](https://github.com/driftyco/ionic-native/blob/71916a85ddbffbaac3967ec3ef45e1fa36c73df0/src/plugins/file.ts#L291) the FileReader object but doesn't fully implement the spec. This issue comment might be helpful: https://github.com/driftyco/ionic-native/issues/505#issuecomment-243434491

[...]

PR #622, opened 2018-06-29
Add workaround for cordova FileReader

vially:
As mentioned in [several](https://github.com/mpetroff/pannellum/issues/284) [other](https://github.com/ionic-team/ionic-native/issues/505) [places](https://github.com/ionic-team/ionic2-app-base/issues/126), [cordova-plugin-file](https://cordova.apache.org/docs/en/latest/reference/cordova-plugin-file/) is [overwriting the FileReader](https://github.com/mpetroff/pannellum/issues/284#issuecomment-260669954) with one that doesn't follow the spec.

This is arguably a problem with `cordova-plugin-file`. But since a fix doesn't seem to be very high on their priority list, maybe adding this temporary workaround for it to `pannellum` would be a good short term solution.

### DIFF
--- a/src/js/pannellum.js
+++ b/src/js/pannellum.js
@@ -518,6 +518,10 @@ function onImageLoad() {
  */
 function parseGPanoXMP(image) {
     var reader = new FileReader();
+    // Fix cordova-plugin-file compatibility
+    if (window.cordova && reader._realReader) {
+        reader = reader._realReader;
+    }
     reader.addEventListener('loadend', function() {
         var img = reader.result;
 


jineshmehta16:
Hi Can this solution be Implemented for normal Ionic framework where pannellum is not included? 

I am facing this issue, I have tried to reorder cordova and polyfill still it didnt solve the problem. In my project panaroma is not required. So how can I address in this situation?

vially:
> Can this solution be Implemented for normal Ionic framework where pannellum is not included?

This solution could not be implemented in Ionic because  it would then break plugins which rely on the implementation of `FileReader` provided by the `cordova-plugin-file` plugin (e.g.: file-system access). The only place where this mess could be fixed properly would be in the `cordova-plugin-file` by either not overwriting the global `FileReader` object or at least overwrite it in a way that is 100% compatible with the spec.

> In my project panaroma is not required. So how can I address in this situation?

Most probably you would have to patch the library that is using the `FileReader` API (most probably in a similar way to my patch here). But obviously that's not a good long-term solution because we can not expect that all the libraries that make use of the `FileReader` API implement a workaround just to be compatible with `cordova-plugin-file`.

mpetroff:
A fix might be low on their priority list, but it's not on mine at all. Mobile / app frameworks are not officially supported; I just added a note to the README to make this policy explicit. While I'll accept pull requests that make minor tweaks to fix compatibility with these frameworks, I'm not adding any code that specifically targets one of them (like this pull request does).

PR #626, opened 2018-07-04
Partial panorama default cube size fix

matejsoukup:
Change computation of cube size - taking haov into account


mpetroff:
Thanks!

Issue #631, opened 2018-07-18
Equirectangular:- Site throws error if it has content security policy because of Blob


vineetkumar1511:
I am trying to load this component in different sites but due to blob used for equirectangular, i get below error:-
> Content Security Policy directive: "script-src https: blob: 'unsafe-inline' 'unsafe-eval'".
This is due to some site has content security policy and they don't allow  below type of url :-

> "blob:https://some-url-path"

Can't we implement equirectangular with blob ? Is there any road map to remove blob ?

mpetroff:
Do you have an example that displays this behavior?

vineetkumar1511:
I do not have test link but below are the steps using which you can reproduce this issue.

1) Open https://www.news.com.au/
2) Open developer console and execute below snippets :-   
> var jsElm = document.createElement("script");
    jsElm.type = "application/javascript";
    jsElm.src = "https://cdnjs.cloudflare.com/ajax/libs/pannellum/2.4.1/pannellum.js";
    document.body.appendChild(jsElm);

> var head  = document.getElementsByTagName('head')[0];
    var link  = document.createElement('link');
    link.id   = "sdfsdf232rw";
    link.rel  = 'stylesheet';
    link.type = 'text/css';
    link.href = 'https://cdnjs.cloudflare.com/ajax/libs/pannellum/2.4.1/pannellum.css';
    link.media = 'all';
    head.appendChild(link);

> var div=document.createElement("div")
div.setAttribute("id","testDiv")
div.style.width = "1000px";
div.style.height = "500px";
document.body.appendChild(div);

> setTimeout(function() {
    pannellum.viewer(document.querySelector("#ad-out-of-page"), {
        type: "equirectangular",
        panorama: "https://massets.bonzai.ad/c2/prod/360Image/360Image_compass_dial.png",
        autoLoad: true,
        compass: true,
        showControls: false,
        hfov: 70
    })
}, 2000);

You can see error after 2 seconds as I am using 2 seconds timeout.

rosemarystanley:
I'm also getting this error.  I tried loading the image with an absolute URL and a relative one. 

```
Refused to load the image 'blob:http://my.localhost/3d8dfcae-46ae-4b48-bc84-3e2224053234' because it violates the following Content Security Policy directive: "default-src 'self' 'unsafe-inline' 'unsafe-eval' data: *". Note that 'img-src' was not explicitly set, so 'default-src' is used as a fallback.
```

[...]

PR #632, opened 2018-07-18
Fix Blob Equirectangular image issue

VineetKumarKushwaha:
We set image src as a blob url for equirectangular image.
This cause content seurity policy issue. As most of the websites only accepts http/https/data for image src.

Fix:
I have passed actual url to parseGPanoXMP function and instead of blob url, I am using actual url.

I have created issue as well.Below is the url :-
https://github.com/mpetroff/pannellum/issues/631


mpetroff:
This isn't a proper fix. See https://github.com/mpetroff/pannellum/issues/631#issuecomment-410100601.

PR #639, opened 2018-08-06
Let customize target attribute for URL HotSpots

Hrumpa:
Sometimes it is needed to open HotSpot link right in the same frame ("_self") or in a named one.


mpetroff:
Thanks!

PR #640, opened 2018-08-07
Changed Math.min to Math.max

chasewalden:
Using min causes odd errors where you can actually "zoom closer" than your minHfov setting. 

If the calculation for the multi-res minHfov is smaller than the config.minHfov, then you can zoom closer. This can happen when cubeResolution is very large and canvas.width is very small. In my case, my cuberesolution is 6704 and  canvas.width is 1440. My minHfov is 30 but the calculated minHfov is around 21, allowing me to zoom-in more than desired.

### DIFF
--- a/src/js/pannellum.js
+++ b/src/js/pannellum.js
@@ -2145,7 +2145,7 @@ function constrainHfov(hfov) {
     // Keep field of view within bounds
     var minHfov = config.minHfov;
     if (config.type == 'multires' && renderer) {
-        minHfov = Math.min(minHfov, renderer.getCanvas().width / (config.multiRes.cubeResolution / 90 * 0.9));
+        minHfov = Math.max(minHfov, renderer.getCanvas().width / (config.multiRes.cubeResolution / 90 * 0.9));
     }
     if (minHfov > config.maxHfov) {
         // Don't change view if bounds don't make sense


mpetroff:
The "zooming closer" is the intended behavior, but it was not documented. The idea was to avoid the need to explicitly set the maximum zoom for multires panoramas. However, as you point out, doing so prevents one from setting the maximum zoom if desired. Thus, I added a boolean `multiResMinHfov` parameter in 35083170d535b8f2fdce4392052aa50c929880dc, which can be set to `true` to cause the `minHfov` value to still be used for multires panoramas.

chasewalden:
Might I propose a suggestion? I believe for sake of consistency and straightforwardness, do not add another config option. Instead, allow the user to set the lower bound for Horizontal FOV using the `minHfov` option, regardless of the panorama being a multires or not. If, however, the `minHfov` option is omitted (`minHfov === undefined`) and the panorama type is multires, then allow the auto-calculation to occur. IMO, I don't think it makes much sense to have two separate options.

An alternative is have a boolean config option, that defaults to true, that allows the `minHfov` value to be overridden by the automatic calculation. If it is false, enforce that the `minHfov` value set by the user is used.

mpetroff:
I agree in principle, but due to the way that the configuration is processed and used, handling the undefined case is not straightforward.

PR #645, opened 2018-08-20
Customizable attributes for Hotspots

DLar:
Sometimes more than just the target attribute needs to be customized on the hotspot. Expansion of ba35ea8

Hotspots now accept an object with a list of attributes to append to the anchor tag.


mpetroff:
Thanks!

PR #675, opened 2018-10-18
Allow friction value to be set in options

willcalderbank:
A client wanted the 'fling' to come to rest quicker increasing the friction value allows this.

I inverted the value (e.g 0 no friction, 1 max friction) as its more intuitive.


willcalderbank:
Apologies, just realised this PR contains lots of commits, let me tidy that up.

PR #689, opened 2018-11-27
Prevent friction value of undefined breaking inertia

willcalderbank:


mpetroff:
I don't see what purpose this serves. The only way the `friction` value is undefined is if one sets it to that, in which case it's just one of many ways a malformed configuration can break things. Currently it breaks things instead of doing something unexpected, which I think is the better failure mode.

willcalderbank:
It's not obvious that an undefined friction value causes the failure, the view just goes blank when it attempts to calculate inertia. Took me 10 mins to figure out what was wrong. This may save someone those 10 minutes. You have similar 'set to default' code else where in the file.

It's up to you, if you'd like to merge it. Don't really mind either way

mpetroff:
I'm not going to merge this. No matter how many checks I add, there are still going to be edge cases that are missed. The existing "set to default" code is for handling optional parameters that aren't set, so while it may accomplish something similar, the purpose is different. Thanks for the pull request anyway.

PR #690, opened 2018-11-27
Add stopMovement to stop all movement

willcalderbank:


mpetroff:
What use case does this serve?

In the future please describe your changes and why you made them; please don't open pull requests that don't have descriptions.

willcalderbank:
Sorry thought the title was description enough. Adds a function stopMovement which stops any movement.

Use case is to stop an animation before it completes. My client wishes to pan view a long distance slowly and have a button to stop movement.Currently in the animate function it check to see if an animation is running and will ignore any subsequent animation calls until the first one has finished so you can't simply animate to current position to stop the first animation. As far as I'm aware currently there is no way to stop an animation while in process, this addresses that.

mpetroff:
Thanks!

PR #707, opened 2019-01-14
Fix webkit webkitIsFullScreen delay

### LINKED ISSUE #629: Fullscreen stretched in Safari
In Safari, the aspect ratio is not set properly when switching to fullscreen mode, causing the viewer to be stretched. It appears that the `onFullScreenChange` method isn’t being called, which means the `fullscreenchange` event isn’t firing.

Browser: Safari Version 11.1.1 (13605.2.8)
Machine: 27" iMac

Copied from comment on: https://mpetroff.net/2018/01/pannellum-2-4/

cleptric:
This fixes #629
It appears to be, that webkit needs a tick to set the `webkitIsFullScreen ` property.

### DIFF
--- a/src/js/pannellum.js
+++ b/src/js/pannellum.js
@@ -2108,19 +2108,21 @@ function toggleFullscreen() {
  * @private
  */
 function onFullScreenChange(resize) {
-    if (document.fullscreenElement || document.fullscreen || document.mozFullScreen || document.webkitIsFullScreen || document.msFullscreenElement) {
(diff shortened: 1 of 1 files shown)


mpetroff:
I'm going to have to think about this some more. Any fix based on timing is inherently fragile, so I'd like to find a better approach.

cleptric:
I agree, but setting a 0 timeout is soley based on ticks

Issue #516, opened 2017-12-07
lookAt rotates panorama to 180 degrees opposite clicked tag


rpcarson:
I can add tags in the expected location, but when calling lookAt, the viewer rotates to the opposite side that the tag is on. I recently updated in order to fix the viewer rotating the wrong direction to lookAt tag as fixed in "Make sure yaw animation is in shortest direction". I was able to fix this new bug by removing the code that was added in "Make sure yaw animation is in shortest direction".

mpetroff:
I assume by "tag" you mean "hot spot," as is used in the documentation.

Are you using the latest Git `master`? I just tested using `lookAt` to point in the direction of a few different hot spots, and it works fine.

rpcarson:
hotspots, sorry.

yes its master. Its used in a webview as part of an ios app.

mpetroff:
Can you put together a short example that shows this bug? As I said, I haven't been able to reproduce it.

gazhay:
I'm experiencing this on the latest version

e.g.

`
console.log("seeking ");
console.log(t.target.getAttribute("data-pitch"),t.target.getAttribute("data-yaw"  ));
Viewer.lookAt( t.target.getAttribute("data-pitch"),t.target.getAttribute("data-yaw"  ), undefined, true, function(){
                        console.log("sought ");
                        console.log(Viewer.getPitch(), Viewer.getYaw());
},{});
`

Results in 

`seeking 
20.68352309466332 74.43418659026379
sought 
20.68352309466332 117.81403106170094
`

gazhay:
There seems to be a serious problem, as sending the same lookAt params twice, returns different "sought" Yaw positions, but the panorama viewer doesn't move.

seeking 
20.68352309466332 74.43418659026379
sought 
20.681291600080698 254.42416729831274
seeking 
20.68352309466332 74.43418659026379
sought 
20.68352309466332 -105.56581340973621seeking 
20.68352309466332 74.43418659026379
sought 
20.681291600080698 254.42416729831274
seeking 
20.68352309466332 74.43418659026379
sought 
20.68352309466332 -105.56581340973621

gazhay:
Also noted - autorotation switches direction after using lookAt.

Which may go some way to explaining why it "looksAt" the wrong direction.

gazhay:
Used Chrome debugger - this line changes the Yaw value in setYaw (line 2474)

    yaw = ((yaw + 180) % 360) - 180 // Keep in bounds

[...]

PR #723, opened 2019-03-23
Fix for issue #516

gazhay:
Although the code works fine in the debugger, google chrome was allowing the commented line to change the value when it should not.

I have subbed the line with a parenthetical and added two parseFloats later in the code as JS was converting yaw value to a string which broke things badly.


mpetroff:
See https://github.com/mpetroff/pannellum/issues/516#issuecomment-475911994.

PR #756, opened 2019-07-21
Extending standalone with more parameters

MasterZydra:
This version of the `standalone.js` supports settings _hotSpotDebug, compass, autoRotateInactivityDelay, autoRotateStopDelay, authorURL, showZoomCtrl and showControls_

### DIFF
--- a/src/standalone/standalone.js
+++ b/src/standalone/standalone.js
@@ -39,8 +39,8 @@ function parseURLParameters() {
             case 'compass':
                 configFromURL[option] = JSON.parse(value);
                 break;
-            case 'author': case 'title': case 'firstScene': case 'fallback':
-            case 'preview': case 'panorama': case 'config':
+            case 'author': case 'authorURL': case 'title': case 'firstScene':
+            case 'fallback': case 'preview': case 'panorama': case 'config':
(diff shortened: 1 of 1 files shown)


mpetroff:
Thanks, but I am hesitant to add any additional URL parameters. You can set all of these in the standalone viewer by using a JSON configuration file.

Unless you can provide a compelling reason why these parameters should be URL parameters instead of requiring a JSON configuration file, I'm going to close this pull request.

MasterZydra:
If there is already the interface to configure the panorama via the URL, there should also be the possibility to make settings like the display and hiding of the compass.  
For some applications, such as SharePoint SPFx webparts, implementation via standalone is easier.

mpetroff:
There is the possibility to set these in the standalone version, via a JSON configuration file:

URL:
`pannellum.htm#config=configuration.json`

`configuration.json`:
```json
{
    "hotSpotDebug": true,
    "compass": false
}
```

MasterZydra:
It is ok if you don't want to add the parameters. For my application case writing in a JSON file was not possible.

mpetroff:
I'm generally not keen on adding additional parameters, so I'm going to close this.

PR #759, opened 2019-07-24
Return JPEG instead of PNG when passing returnImage option to render

willcalderbank:
Its significantly faster to create a jpeg image than a png. This speeds up the time before transitions between scenes when fades are used.

### DIFF
--- a/src/js/libpannellum.js
+++ b/src/js/libpannellum.js
@@ -738,7 +738,7 @@ function Renderer(container) {
         }
         
         if (params.returnImage !== undefined) {
-            return canvas.toDataURL('image/png');
+            return canvas.toDataURL('image/jpeg');
         }
     };
     


strarsis:
Some devs, including me, on the other hand, need lossless PNG for further processing, 
it would be better to offer an argument for the image type (PNG or JPEG).

mpetroff:
I don't think this is the correct approach. I never noticed a slow down for transitions before, but when I went looking for it, I could reproduce it, but only on a 4k display while full screen. I think the better approach is to skip the image encoding entirely and instead use an [`ImageBitmap`](https://developer.mozilla.org/en-US/docs/Web/API/ImageBitmap) to copy to another `<canvas>` element.

willcalderbank:
I also never noticed the slow down until a client ran on an underpowered machine, then the effects were fairly noticeable. I went in to see if I could speed it up in any way and chromes debugging tools told me that the rendering of the image is by far the slowest part. PNG took on average 400ms while PNGs took 100ms-ish.

Im a bit snowed under at the moment but when I get a few seconds ill look into creating another canvas. I did quickly look at rendering it to an offscreen canvas but gave up as soon as I discovered pngs where fast enough.

mpetroff:
I have good idea of the changes that need to be made. If I have time, I'll try to make the changes sometime in the next few weeks.

5ko:
We can check if the params.returnImage value is an image mime type, and if so, return that type, otherwise return PNG like before:
```javascript
        if (params.returnImage !== undefined) {
            if (params.returnImage.toString().indexOf('image/') == 0)
                return canvas.toDataURL(params.returnImage);
            else  return canvas.toDataURL('image/png');
        }
```
This will not break existing installations that rely on the PNG type.

mpetroff:
I finally got around to working on this. I switched to using `ImageBitmap` for fades and added support for specifying `image/jpeg` or `image/png` as alternatives for the returned image in a7a8b85c34d8551546fc5c34b99bc1623774cd45. If `returnImage` is set to `true` (it used to be a Boolean value), PNG is still used to avoid breaking older code.

PR #760, opened 2019-07-24
Add scale option to hotspot

willcalderbank:
When `true`, the hotspot is scaled to match the fish eye effect of the image.

Hotspots then appear to be embedded in the panorama instead of floating on top


mpetroff:
Thanks!

Issue #744, opened 2019-06-14
Loading a scene causes autorotate pan to change in an unexpected way


willcalderbank:
If you load a scene then the auto rotate kicks in (or you call `startAutoRotate()`) the viewer pans to the pitch of the scene was loaded with.

This occurs if you load the scene manually with `loadScene()` or use the scene hotspot type. 

The problem is in `init()` where `origPitch = config.pitch;`. When `loadScene()` is called `config.pitch` is set `load()` is then called which calls `init()`

`origPitch` is then used to change pitch before auto rotating.

mpetroff:
That's the intended behavior.

mpetroff:
See 5228f48f3087e75227ee37aa6b4cf074d7ea97ac and discussion in #170.

willcalderbank:
In that case should the documentation mention it? It may be the intended behavior but it still unexpected that loadScene has such a side effect.

mpetroff:
Now I'm no longer sure we're talking about the same thing, since the behavior I'm referring to has nothing to do with `loadScene`. When auto-rotation starts, it always pans to the initial pitch; this is true regardless of how the scene is loaded, e.g., this is also true for the initial panorama.

I added some additional documentation in d02cedcc4e8671b9d76bac473cb94a85bfbd10b5.

willcalderbank:
Im talking about `loadScene`, it calls `load()` which intern calls `init()`. So yes, no matter how the scene is is loaded be that from the initial panorama to `loadScene` the initial pitch is set. What I found unexpected is how `loadScene` also sets the initial pitch. In the coming days I will make a PR with a setInitialView or possibly a startAutoPan with a pitch to pan to first param.  A client wishes to have tour but the initial view of a scene is nowhere near the horizon and they want the autopan to kick in to view the said horizon. 

Thanks

mpetroff:
Okay, that makes sense now. I'll reopen this issue and wait for the pull request.

mpetroff:
Implemented in #761.

PR #761, opened 2019-07-24
Add optional pitch argument to startAutoRotate

willcalderbank:
Before starting rotation, the viewer is panned to `pitch`.

If pitch is not specified, origPitch will be used

Addresses #744


mpetroff:
Thanks!

PR #777, opened 2019-08-07
Adding Automated Testing with CircleCI

vsoch:
This will add automated testing, which I was able to get working with CircleCI. Specifically:

 - The testing script has an argument parser for easier use
 - Screen shots don't seem to work with an element, but instead we can take them with the driver, and then use PIL to crop them appropriately.
 - We use the same method to calculate differences.

You'll need to connect CircleCI (and you can disconnect Travis) to get the tests up and running! You can look at my branch tests [here](https://circleci.com/gh/vsoch/pannellum/25) to see tests working. I've also done some tweaks to the README that I suggested in the JoSS review, since it was easy enough anyway!

Please feel free to edit this PR as  you see fit! We are close to checking off that Automated Tests checkbox for the review, I think :)


vsoch:
I'd also like to comment how incredibly *awesome* it is that this hit issue 777! It must be good luck! :D

mpetroff:
Thank you so much! This is above and beyond what I would have expected from a reviewer.

I applied this in a squashed commit in 13ceeb605601776a57c69f1ba00be70e5d968735, since I wanted to completely remove the driver binaries from the repository history. As the drivers are both platform and browser version specific, I don't think it makes sense to include them in the repository.

I was able to get the element screenshots working again the way they were working in Firefox by updating the ChromeDriver version; per the [changelog](https://chromedriver.chromium.org/downloads), the command was fixed in version 2.42.

Using Chrome, I was able to get Travis to work, so I stuck with it over CircleCI, since I'm more familiar with Travis. However, if I ever decide to switch, I have a starting point, thanks to you.

PR #779, opened 2019-08-09
Adding complete example with Dockerfile to use generate.py

vsoch:
This pull request will provide a simple Dockerfile for the user to build a container with all required dependencies to run generate.py. It's likely that a new user won't have hugin, and perhaps some of the Python packages, and the container will get them up and running much more quickly than figuring out the dependencies. I've also written up a README with full instructions for using generate!

Signed-off-by: Vanessa Sochat <vsochat@stanford.edu>

### DIFF
--- a/utils/multires/Dockerfile
+++ b/utils/multires/Dockerfile
@@ -1,0 +1,19 @@
+FROM ubuntu:18.04
+
+# docker build -t generate-panorama .
+# docker run -it -v $PWD:/data generate-panorama /data/image.png
+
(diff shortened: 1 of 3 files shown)


mpetroff:
While I don't use Docker personally, this looks like a useful addition. Thanks!

Since the script's dependencies are stable, mature packages, I simplified the Dockerfile to just use packages from the default repos. I also rewrote parts of the readme such that it's no longer just about Docker.

vsoch:
Sounds good! I'm giving the new recipe one final test to make sure things are working as they should.

PR #813, opened 2019-11-29
Use actual image URL when blob fails to load

sayali-15:
Most of the websites accept http/https/data for image src.
Since we are setting image src as blob, this causes issues.

Previously a [solution](https://github.com/mpetroff/pannellum/pull/632) was proposed to set the source to actual image URL.
As per [this comment](https://github.com/mpetroff/pannellum/issues/631#issuecomment-410100601),
I have parsed csp header to check if the site blocks blob. If so, we fall back to actual URL.

### DIFF
--- a/src/js/pannellum.js
+++ b/src/js/pannellum.js
@@ -395,7 +395,7 @@ function init() {
                     anError(config.strings.fileAccessError.replace('%s', a.outerHTML));
                 }
                 var img = this.response;
-                parseGPanoXMP(img);
+                parseGPanoXMP(img, p);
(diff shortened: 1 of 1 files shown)


mpetroff:
Thanks! I merged this and then make some minor changes to it in 15e0d09fec8d7ea56adb67f30736c81d46f1aa88.

In the future, please make pull requests off of feature branches; you made a feature branch (`cspfix`), but you merged it into your fork's copy of `master` and opened the pull request using that instead. If the pull request was made using the feature branch, I would have been able to make the changes I made after merging the pull request before merging it, which I would have preferred.

PR #846, opened 2020-03-26
Fix missing DeviceMotionEvent error in Safari (desktop)

maxmishyn:
Desktop Safari (at least 11.x) is failing to open Pannellum due to the JS error "Can't find variable: DeviceMotionEvent". This PR fixes above error

### DIFF
--- a/src/js/pannellum.js
+++ b/src/js/pannellum.js
@@ -2369,7 +2369,8 @@ function stopOrientation() {
  * @private
  */
 function startOrientation() {
-    if (typeof DeviceMotionEvent.requestPermission === 'function') {
+    if (typeof DeviceMotionEvent === 'function' &&
+        typeof DeviceMotionEvent.requestPermission === 'function') {
         DeviceOrientationEvent.requestPermission().then(function(response) {
             if (response == 'granted') {
                 orientation = 1;


mpetroff:
Apparently Apple can't even make Safari self-consistent. The code in question was added because of the iOS version of Safari.

I assume you had `orientationOnByDefault` set to `true`, since I don't see any other way that code could have executed.

I made a similar fix in 9592d2ab7be749faf6f590b8c6034f42ccd3d0d9. While browser implementations seem to all return `'function'` for `typeof DeviceMotionEvent`, the spec defines it as an interface, so I'm not sure that behavior is guaranteed. Thus, I relaxed the check to only make sure `DeviceMotionEvent` is defined instead. I also moved the orientation support check to also be enforced for  `orientationOnByDefault`.

maxmishyn:
Yes you are right. In my usage scenario orientationOnByDefault is true. 
Moreover I not 100% sure, but I do not remember that I experienced same issue with older versions of desktop Safari.
You fix in 9592d2a  looks more complete. It is strange that I missed it when I was looking through known issues ;)

mpetroff:
You didn't miss it. I made it based on this pull request; I apologize for not being more clear about that.

PR #861, opened 2020-05-08
Add touchend event to make clickHandlerFunc work on touch devices

pfortes:
Using a click handler in a hotspot with clickHandlerFunc and clickHandlerArgs didn't work on touch devices, adding the touchend event fixes this.

### DIFF
--- a/src/js/pannellum.js
+++ b/src/js/pannellum.js
@@ -1800,6 +1800,9 @@ function createHotSpot(hs) {
         div.addEventListener('click', function(e) {
             hs.clickHandlerFunc(e, hs.clickHandlerArgs);
         }, 'false');
+        div.addEventListener('touchend', function(e) {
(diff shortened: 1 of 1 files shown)


mpetroff:
Thanks! This pull request is an incomplete solution since it doesn't handle touch input in browsers that use pointer events instead of touch events, but I implemented the more complete solution in 6075769935a27e61c9e1f0f656f8cef532abaa77.

pfortes:
Missed that detail.
Thank you! 
And thanks for all your hard work in this great library, hope i can contribute more in the future.

pfortes:
@mpetroff i have one question, do changes like this one get pushed to npm or should i use a local copy of the newest build instead?

mpetroff:
Changes only get pushed to npm when there's a release. Until then, you'll need to use a local copy.

PR #880, opened 2020-06-13
Feature offer

NiHoel:
Implementing my tour editor, I added quite some features to pannellum. I opened this pull request to review the features and discuss which of them would be worth adding to the master. Having done this, I would prepare a new pull request that only contains these features and close this one.

Pannellum in action with the new features: [https://nihoel.github.io/cirspecte/edit.html?tour=https%3A%2F%2Fnihoel.github.io%2Fcirspecte-demo%2Ftour.json](https://nihoel.github.io/cirspecte/edit.html?tour=https%3A%2F%2Fnihoel.github.io%2Fcirspecte-demo%2Ftour.json) (open the panorama tab to drag hotspots)

The new features are:
* A new multiresolution panoram type (multiresrec) based on a tiled equirectangular base image
* Option to interpolate between tiles (only for multiresrec at the moment)
* Hotspot dragging (hotspot attribute draggable + dragHandler function)
* External tile loading (allows to pass a function that loads tiles - useful when accessing local files via electron or cordova)


mpetroff:
You're demo link is broken and does not work for me in either Firefox or Chrome, with multiple errors reported in the console.

> A new multiresolution panoram type (multiresrec) based on a tiled equirectangular base image

Equirectangular is a compromise projection, the only advantage of which is that it is extremely simple. It has both extreme area distortions and extreme angle distortions, which result in excessively large file sizes and sampling issues. As the addition complexities of many better projections aren't a problem, the only reason to use the equirectangular projection is that it is the de facto standard interchange format for panoramic images (which is why Pannellum supports it). However, as soon as you need to do pre-processing steps to prepare the image, such as to use it as a tiled multiresolution format, it loses its only advantage. Thus, I will _never_ include a multiresolution format based on equirectangular images in Pannellum. I have plans to eventually implement a new multiresolution format, but it will use a low angular distortion equal-area projection, which is quantitatively better than the existing multiresolution format, not quantitatively worse.

> Option to interpolate between tiles (only for multiresrec at the moment)

What advantage does this have? I don't see any reason why one would need to interpolate between tiles, since there aren't any gaps between tiles. Am I misunderstanding what this does?

> Hotspot dragging (hotspot attribute draggable + dragHandler function)

Depending on the implementation, this could be interesting. However, I can't reach a conclusion about this without first seeing how it works (as I said, your demo is broken) and seeing the implementation (I'm not about to dig through >1k lines and dozens of commits to try to figure out what changes are what).

> External tile loading (allows to pass a function that loads tiles - useful when accessing local files via electron or cordova)

As is stated in the readme, web browsers are the only supported target platform. If you want to change how tiles are loaded to support something else, you can always monkey-patch `XMLHttpRequest` without needing to change Pannellum, or better yet, get the framework you're using to fix their implementation to make it standards-compliant. While I'm not enthusiastic about such a feature, it's something I might merge, depending on the implementation (see previous comment about the length of your pull request).

NiHoel:
> 
> 
> You're demo link is broken and does not work for me in either Firefox or Chrome, with multiple errors reported in the console.
> 
Hm, I have no idea why it works for me on Firefox, Chrome and Edge but not for you on any of these browsers. I might have a look at that tomorrow.


> > A new multiresolution panoram type (multiresrec) based on a tiled equirectangular base image
> 
> Equirectangular is a compromise projection, the only advantage of which is that it is extremely simple. It has both extreme area distortions and extreme angle distortions, which result in excessively large file sizes and sampling issues. As the addition complexities of many better projections aren't a problem, the only reason to use the equirectangular projection is that it is the de facto standard interchange format for panoramic images (which is why Pannellum supports it). However, as soon as you need to do pre-processing steps to prepare the image, such as to use it as a tiled multiresolution format, it loses its only advantage. Thus, I will _never_ include a multiresolution format based on equirectangular images in Pannellum. I have plans to eventually implement a new multiresolution format, but it will use a low angular distortion equal-area projection, which is quantitatively better than the existing multiresolution format, not quantitatively worse.
> 

A fair point.

> > Option to interpolate between tiles (only for multiresrec at the moment)
> 
> What advantage does this have? I don't see any reason why one would need to interpolate between tiles, since there aren't any gaps between tiles. Am I misunderstanding what this does?
> 

I mean the following effect (left: default; right: with interpolation):
tileEdge

It only happens when displayed resolution is higher than the panorama can provide for this area.

> > Hotspot dragging (hotspot attribute draggable + dragHandler function)
> 
> Depending on the implementation, this could be interesting. However, I can't reach a conclusion about this without first seeing how it works (as I said, your demo is broken) and seeing the implementation (I'm not about to dig through >1k lines and dozens of commits to try to figure out what changes are what).
> 
> > External tile loading (allows to pass a function that loads tiles - useful when accessing local files via electron or cordova)
> 
> As is stated in the readme, web browsers are the only supported target platform. If you want to change how tiles are loaded to support something else, you can always monkey-patch `XMLHttpRequest` without needing to change Pannellum, or better yet, get the framework you're using to fix their implementation to make it standards-compliant. While I'm not enthusiastic about such a feature, it's something I might merge, depending on the implementation (see previous comment about the length of your pull request).

Discarding the equirectangular stuff, the changes will become much less. I will update the PR in the next days.

NiHoel:
Suspended in favor of #881

Issue #478, opened 2017-09-17
Turning orientationOnByDefault with javascript


DStillingfleet:
Hi Matthew

Thank you for your great work.

I'm using API to load an pannellum instance, and will set the orientationOnByDefault to false so that I can manipulate the panorama with lookAt, setYaw, etc.  When these actions are complete, I wish to set the orientationOnByDefault to true so that the panorama will pan / tilt with the orientation of the mobile device.  

Can you provide some guidance please.

Kind regards

Derek

DStillingfleet:
P.S.

Is there a way to set the direction of rotation with lookAt?

mpetroff:
This was literally just asked for in #477, so I added `isOrientationSupported()`, `startOrientation()`, and `stopOrientation()` functions to the API in 35edd9f5f7d7900515ee77d155c7d9a7683203c8.

The `lookAt` rotates in the direction that's shortest; this is not configurable (there's a bug in 2.3.2 for the rotation direction, that was fixed in 129bc0310d71a1d483c4bdbcf01c266fde98d767).

DStillingfleet:
Thank you Matthew.

This will be the first time I have tried to downloads a development version from GitHub.  

I've copied across pannellum.css, RequestAnimationFrame.js, libpannellum.js, pannellum.js, and pannellum.htm from the src directories, but when I try my webpage pannellum is stuck in loading the panorama.  What am I missing?

The bug fix for the shortest direction of rotation in LookAt is exactly what I was hoping to achieve.

Best wishes


Derek

mpetroff:
What does your browser's developer console say?

DStillingfleet:
The developer console reports...

ReferenceError: libpannellum is not defined      Pannellum.js:430:9

DanielBiegler:
Are you loading the scripts in the right order? Like so:

```
.
.
<script type="text/javascript" src="../js/libpannellum.js"></script>
<script type="text/javascript" src="../js/RequestAnimationFrame.js"></script>
<script type="text/javascript" src="../js/pannellum.js"></script>
<script type="text/javascript" src="standalone.js"></script>
.
.
```

DStillingfleet:
Ah, I didn't realise I had to load the scripts myself. Had assumed that pannellum.htm did that.  I'll give it a try.

Thank you.

mpetroff:
[...]

PR #881, opened 2020-06-15
Feature offer

NiHoel:
This pull request superseds #880 

Pannellum in action with the new features: https://nihoel.github.io/cirspecte/edit.html?tour=https%3A%2F%2Fnihoel.github.io%2Fcirspecte-demo%2Ftour.json (open the panorama tab to drag hotspots)

The new features are:
* Hotspot dragging (hotspot attribute draggable + dragHandler function)
* External tile loading (allows to pass a function that loads tiles - hooked in at the processNextTile level; passes internal node object to caller; https://github.com/NiHoel/pannellum/blob/178a08c776f99ef270fc2ee564622e6a04265dd3/src/js/libpannellum.js#L1182-L1193)

* Accept images as parameter (https://github.com/NiHoel/pannellum/blob/178a08c776f99ef270fc2ee564622e6a04265dd3/src/js/pannellum.js#L386-L389)
* Option orientationAlignNorth - use absolute orientation values if enabled; basically switches between the implementation before and after commit f0fcf31998aa149212546dd07276bd93bc01a01c (https://github.com/NiHoel/pannellum/blob/178a08c776f99ef270fc2ee564622e6a04265dd3/src/js/pannellum.js#L1659) - 
* Add getters and setters for orientationOnByDefault (https://github.com/NiHoel/pannellum/blob/178a08c776f99ef270fc2ee564622e6a04265dd3/src/js/pannellum.js#L2943-L2946)
* Enable orientationSupport for android regardless of the protocol (https://github.com/NiHoel/pannellum/blob/178a08c776f99ef270fc2ee564622e6a04265dd3/src/js/pannellum.js#L262-L263) - otherwise file-like locations do not work

The last four things I noticed when going through my commits. Tell me, if I should also remove them.

### DIFF
--- a/src/js/libpannellum.js
+++ b/src/js/libpannellum.js
@@ -865,12 +865,7 @@ function Renderer(container) {
         this.level = level;
         this.x = x;
         this.y = y;
-        if (!path) {
-            this.path = side + '_' + level + '_' + x + '_' + y;
-        } else {
-            this.path = path.replace('%s', side).replace('%l', level).replace('%x', x).replace('%y', y);
-        }
-        this.uri = encodeURI(this.path + '.' + image.extension);
+        this.path = path.replace('%s',side).replace('%l',level).replace('%x',x).replace('%y',y);
     }
 
     /**
@@ -1132,16 +1127,12 @@ function Renderer(container) {
     /**
      * Processes a loaded texture image into a WebGL texture.
      * @private
-     * @param {Image | ImageBitmap | ImageData | HTMLCanvasElement} img - Input image.
+     * @param {Image} img - Input image.
      * @param {WebGLTexture} tex - Texture to bind image to.
      */
     function processLoadedTexture(img, tex) {
         gl.bindTexture(gl.TEXTURE_2D, tex);
-        if (img instanceof HTMLCanvasElement) {
-            var data = img.getContext('2d').getImageData(0, 0, img.width, img.height);
-            gl.texImage2D(gl.TEXTURE_2D, 0, gl.RGB, gl.RGB, gl.UNSIGNED_BYTE, data)
-        }else
-            gl.texImage2D(gl.TEXTURE_2D, 0, gl.RGB, gl.RGB, gl.UNSIGNED_BYTE, img);
+        gl.texImage2D(gl.TEXTURE_2D, 0, gl.RGB, gl.RGB, gl.UNSIGNED_BYTE, img);
         gl.texParameteri(gl.TEXTURE_2D, gl.TEXTURE_MAG_FILTER, gl.LINEAR);
         gl.texParameteri(gl.TEXTURE_2D, gl.TEXTURE_MIN_FILTER, gl.LINEAR);
         gl.texParameteri(gl.TEXTURE_2D, gl.TEXTURE_WRAP_S, gl.CLAMP_TO_EDGE);
@@ -1175,25 +1166,10 @@ function Renderer(container) {
             this.image.addEventListener('error', loadFn); // ignore missing tile file to support partial image, otherwise retry loop causes high CPU load
         }
 
-        TextureImageLoader.prototype.loadTexture = function(node, src, texture, callback) {
+        TextureImageLoader.prototype.loadTexture = function(src, texture, callback) {
             this.texture = texture;
             this.callback = callback;
-            if (src instanceof Function) {
-                src(JSON.parse(JSON.stringify(node)), this.image, this.texture).then(img => {
-                    if (!img)
-                        this.callback(this.texture, false);
-                    else if (img != this.image) {
-                        processLoadedTexture(img, this.texture);
-                        this.callback(this.texture, true);
-                    }
-                    releaseTextureImageLoader(this);
-                }).catch(() => {
-                    this.callback(this.texture, false);
-                    releaseTextureImageLoader(this);
-                });
-            } else {
-                this.image.src = src;
-            }
+            this.image.src = src;
         };
 
         function PendingTextureRequest(node, src, texture, callback) {
@@ -1206,7 +1182,7 @@ function Renderer(container) {
         function releaseTextureImageLoader(til) {
             if (pendingTextureRequests.length) {
                 var req = pendingTextureRequests.shift();
-                til.loadTexture(req.node, req.src, req.texture, req.callback);
+                til.loadTexture(req.src, req.texture, req.callback);
             } else
                 textureImageCache[cacheTop++] = til;
         }
@@ -1218,7 +1194,7 @@ function Renderer(container) {
             crossOrigin = _crossOrigin;
             var texture = gl.createTexture();
             if (cacheTop)
-                textureImageCache[--cacheTop].loadTexture(node, src, texture, callback);
+                textureImageCache[--cacheTop].loadTexture(src, texture, callback);
             else
                 pendingTextureRequests.push(new PendingTextureRequest(node, src, texture, callback));
             return texture;
@@ -1231,7 +1207,7 @@ function Renderer(container) {
      * @param {MultiresNode} node - Input node.
      */
     function processNextTile(node) {
-        loadTexture(node, image.loader || node.path + '.' + image.extension, function (texture, loaded) {
+        loadTexture(node, node.path + '.' + image.extension, function(texture, loaded) {
             node.texture = texture;
             node.textureLoaded = loaded ? 2 : 1;
         }, globalParams.crossOrigin);
--- a/src/js/libpannellum.js
+++ b/src/js/libpannellum.js
@@ -865,7 +865,12 @@ function Renderer(container) {
         this.level = level;
         this.x = x;
         this.y = y;
-        this.path = path.replace('%s',side).replace('%l',level).replace('%x',x).replace('%y',y);
+        if (!path) {
+            this.path = side + '_' + level + '_' + x + '_' + y;
+        } else {
+            this.path = path.replace('%s', side).replace('%l', level).replace('%x', x).replace('%y', y);
+        }
+        this.uri = encodeURI(this.path + '.' + image.extension);
     }
 
     /**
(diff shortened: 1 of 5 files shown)


NiHoel:
I'll have a look at the failed test tomorrow: https://travis-ci.org/github/mpetroff/pannellum/builds/698704538

mpetroff:
I figured out why your link wasn't working for me; you managed to unintentionally build a website that doesn't work on Linux. The site has an error at line 1818 of `filesystem.js`, since `platform.os.version` is `null`. I need to add a break point on that line, run `platform.os.version = '1'` in the console while code execution is paused, and then continue code execution to get it to work.

Also, thanks for pointing out what I was missing about the tile interpolation. I guess I'd describe it as "interpolation across tile boundaries" instead of "interpolation between tiles," since "between" made me think of a gap between tiles.

As a general comment on the code, you can't use ES6 as this breaks the build script (and older browser support). Eventually, I'll updated this, probably for Pannellum 3.0, but, until then, you need to stick with ES5.

I don't see anything obvious that would cause the CI test to fail, but I can take a closer look if it's still failing after the point-by-point comments below are addressed.

> Hotspot dragging (hotspot attribute draggable + dragHandler function)

This looks like a good addition! I'll do a line-by-line code review after other issues are addressed.

> External tile loading (allows to pass a function that loads tiles - hooked in at the processNextTile level; passes internal node object to caller; https://github.com/NiHoel/pannellum/blob/178a08c776f99ef270fc2ee564622e6a04265dd3/src/js/libpannellum.js#L1182-L1193)

I'm not sure I see the purpose of this. I don't see why this would need to be done unless you're trying to load local files; however, if you're using local files (unless they're massive), there's no need to use the multiresolution functionality. I'm hesitant to include this, particularly since it will need to be rewritten for ES5-compatibility.

> Accept images as parameter (https://github.com/NiHoel/pannellum/blob/178a08c776f99ef270fc2ee564622e6a04265dd3/src/js/pannellum.js#L386-L389)

This also looks like a good addition.

> Option orientationAlignNorth - use absolute orientation values if enabled; basically switches between the implementation before and after commit [f0fcf31](https://github.com/mpetroff/pannellum/commit/f0fcf31998aa149212546dd07276bd93bc01a01c) (https://github.com/NiHoel/pannellum/blob/178a08c776f99ef270fc2ee564622e6a04265dd3/src/js/pannellum.js#L1659) -

This change was made to fix a bug that caused issues when turning device orientation mode on and off in Safari (see https://github.com/mpetroff/pannellum/issues/478#issuecomment-333000677). I haven't explicitly tested it, but, by looking at the code, I think this change will reintroduce the bug. It also doesn't actually use absolute orientation values in all browsers.

> Add getters and setters for orientationOnByDefault (https://github.com/NiHoel/pannellum/blob/178a08c776f99ef270fc2ee564622e6a04265dd3/src/js/pannellum.js#L2943-L2946)

I'm not sure I understand the purpose of this, since it's a parameter that's only used when the image is first loaded.

> Enable orientationSupport for android regardless of the protocol (https://github.com/NiHoel/pannellum/blob/178a08c776f99ef270fc2ee564622e6a04265dd3/src/js/pannellum.js#L262-L263) - otherwise file-like locations do not work

This is a regression. It will enable orientation support for HTTP, which doesn't work.

mpetroff:
Ignore my comment about the CI failure. It looks like you already fixed that.

NiHoel:
> 
> 
> I figured out why your link wasn't working for me; you managed to unintentionally build a website that doesn't work on Linux. The site has an error at line 1818 of `filesystem.js`, since `platform.os.version` is `null`. I need to add a break point on that line, run `platform.os.version = '1'` in the console while code execution is paused, and then continue code execution to get it to work.

Thanks for investigating the issue with Linux. I fixed it, so it should work for you as well. I totally forgot that this code might cause an issue on non-windows platforms. Unfortunately, I found no better way to reliably detect these features apart from browser specific tests. 
> 
> Also, thanks for pointing out what I was missing about the tile interpolation. I guess I'd describe it as "interpolation across tile boundaries" instead of "interpolation between tiles," since "between" made me think of a gap between tiles.

Since I implemented it only for multiresrec (equirectangular base image) at the moment, I'd only add it for multires, if you really want to have this feature.
> 
> As a general comment on the code, you can't use ES6 as this breaks the build script (and older browser support). Eventually, I'll updated this, probably for Pannellum 3.0, but, until then, you need to stick with ES5.

I have removed the ES6 related code. Regarding the loader code, it is syntactically correct in ES5 and - in my opinion - it's the caller's responsibility to care about backward compatibility or not. In both cases the passed function can return a then-able and catch-able object that runs fine. 

> 
> I'm not sure I see the purpose of this. I don't see why this would need to be done unless you're trying to load local files; however, if you're using local files (unless they're massive), there's no need to use the multiresolution functionality. I'm hesitant to include this, particularly since it will need to be rewritten for ES5-compatibility.

It makes sense if you do not have full control how tiles are served by the server (e.g. no file endings, user authentication, ...)


> > Add getters and setters for orientationOnByDefault (https://github.com/NiHoel/pannellum/blob/178a08c776f99ef270fc2ee564622e6a04265dd3/src/js/pannellum.js#L2943-L2946)
> 
> I'm not sure I understand the purpose of this, since it's a parameter that's only used when the image is first loaded.

I have the issue that orientationOnByDefault is applied when switching between scenes. loadScene() -> processOptions() -> startOrientation(). I haven't tested yet whether this occurs when scenes are loaded by clicking a hotspot. In my case, I dynamically add every scene and immidiately load it by calling loadScene().

tests/tests.html
```diff
@@ -90,7 +90,51 @@
...
+					}
+					img.onerror = function(err) {
+						if(rejector)
+							rejector(err);
+					}
+					
+					img.src = node.uri;
+					
+					return promise;
+				}
```
NiHoel, line 135: A lightweight ES5 compatible implementation of the loader function (race conditions may occur, no chaining possible).

NiHoel:
A few considerations regarding the dragHandlerFunc:
* When to call: start, move, end? (move will require some additional code)
* Pass hot spot so that caller has easily access to pitch and yaw?
* Pass dragHandlerArgs to identify hot spot - like clickHandlerFunc does?

src/js/libpannellum.js
```diff
@@ -1127,12 +1132,16 @@ function Renderer(container) {
...
      * @param {WebGLTexture} tex - Texture to bind image to.
      */
     function processLoadedTexture(img, tex) {
         gl.bindTexture(gl.TEXTURE_2D, tex);
-        gl.texImage2D(gl.TEXTURE_2D, 0, gl.RGB, gl.RGB, gl.UNSIGNED_BYTE, img);
+        if (img instanceof HTMLCanvasElement) {
+            var data = img.getContext('2d').getImageData(0, 0, img.width, img.height);
+            gl.texImage2D(gl.TEXTURE_2D, 0, gl.RGB, gl.RGB, gl.UNSIGNED_BYTE, data)
+        }else
+            gl.texImage2D(gl.TEXTURE_2D, 0, gl.RGB, gl.RGB, gl.UNSIGNED_BYTE, img);
```
mpetroff, line 1144: I'm not fond of adding a special case just for a canvas element. The caller can use `getImageData` externally and just pass the `ImageData` instead.

src/js/pannellum.js
```diff
@@ -378,8 +380,14 @@ function init() {
         
         if (config.dynamic !== true) {
             // Still image
+            if (config.panorama instanceof Image) {
```
mpetroff, line 383: Also check for instance of `ImageData` or `ImageBitmap` here?

src/js/pannellum.js
```diff
@@ -1412,7 +1438,8 @@ function animate() {
     } else if (renderer && (renderer.isLoading() || (config.dynamic === true && update))) {
         requestAnimationFrame(animate);
     } else {
-        fireEvent('animatefinished', {pitch: _this.getPitch(), yaw: _this.getYaw(), hfov: _this.getHfov()});
+        if (_this.getPitch && _this.getYaw && _this.getHfov)
```
mpetroff, line 1441: Why is this check necessary?

src/js/libpannellum.js
NiHoel: Will remove that

src/js/pannellum.js
NiHoel: Yes, with the following steps I can reproduce an error:
1. Use the source files of the viewer
2. Instantiate a viewer (from javascript)
   * Pass a first scene
   * Set autoLoad = true of the scene
   * The scene type is equirectangular
3. The call stack is the following
   viewer -> Viewer -> processOptions -> init -> onImageLoad -> renderInit -> init  -> renderInitCallback -> animateInit -> animate 
4. Line 1438 will evaluate to false (due to type equirectangular) but `_this` has not yet the functions getPitch() ... registered.
5. Line 1442 fails to execute

NiHoel:
> 
> I'd include `dragHandlerArgs` to match `clickHandlerArgs`, since this is consistent and allows the most flexibility.
> 
Added.
> I'm not sure what you mean by your question "When to call: start, move, end?".

When you listen to normal drag events. There is a `dragstart`, `dragmove`, and `dragend` event. At the moment, dragHandlerFunc will only be called when dragging is started and ended. Moreover, the event type is something like `pointerdown` or `pointerup`. I could not use the drag events of the hotspot because dragging ended when the mouse left the hotspot div (which already  happens when you move your mouse with normal speed).

src/js/pannellum.js
NiHoel: Done.

src/js/pannellum.js
mpetroff: Okay, that makes sense. It's a configuration I've never tested.

mpetroff:
Thanks for working with me on this (and putting up with all the changes I wanted)! The hot spot dragging is a nice feature that I anticipate that others will find useful in the future, so thank you for contributing it. After making a few minor formatting changes, I merged the pull request.

PR #883, opened 2020-06-16
startAutoRotate | possible improvement

praefectius:
`startAutoRotate(speed, pitch, hfov)`

Use case:

1. by using "getPitch()" and "getHfov()" as parameter you can just rotate your panorama with the current view.
2. if origHfov is set to a small value (eg. to start with a highlight inside a big panorama) you still can use startAutoRotate with a wider view


mpetroff:
Looks good! Thanks!

PR #918, opened 2020-09-29
Add support for initial view settings

Elbandi:
Based on photosphere doc:
https://developers.google.com/streetview/spherical-metadata

### DIFF
--- a/src/js/pannellum.js
+++ b/src/js/pannellum.js
@@ -577,7 +577,10 @@ function parseGPanoXMP(image, url) {
                 topPixels: getTag('GPano:CroppedAreaTopPixels'),
                 heading: getTag('GPano:PoseHeadingDegrees'),
                 horizonPitch: getTag('GPano:PosePitchDegrees'),
-                horizonRoll: getTag('GPano:PoseRollDegrees')
+                horizonRoll: getTag('GPano:PoseRollDegrees'),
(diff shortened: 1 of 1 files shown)


src/js/pannellum.js
```diff
@@ -605,7 +608,14 @@ function parseGPanoXMP(image, url) {
                         config.horizonRoll = xmp.horizonRoll;
                 }
                 
-                // TODO: add support for initial view settings
+                if (xml.pitch != null && xml.yaw != null && xml.hfov != null) {
```
mpetroff, line 611: This should be `xmp`. Also, I see no reason why all three should need to be defined; some of the other parameters are grouped because they interact with each other, but these parameter do not. I'd use something like this for each of the three parameters:
```javascript
if (xmp.pitch !== null && specifiedPhotoSphereExcludes.indexOf('pitch') < 0)
    config.pitch = xmp.pitch;
```

mpetroff (changes requested):
Have you tested these additions with actual images?

src/js/pannellum.js
Elbandi: typo fixed in forced push. it's doesnt make sense, if someone sets only one parameter (pitch or yaw or hfov). But if you like, i can split it.

mpetroff:
Thanks!

PR #919, opened 2020-09-29
Dockerfile tweak

Elbandi:
- Upgrade ubuntu to newer apps.
- Delete apt lists files ([best practice](https://github.com/docker-library/golang/blob/master/Dockerfile-debian.template#L4-L10))


mpetroff:
According to the [Docker best practices documentation](https://docs.docker.com/develop/develop-images/dockerfile_best-practices/#apt-get), explicitly clearing the APT cache isn't needed for the official Ubuntu images (as are used here), but it doesn't hurt to include it. Thanks!

Issue #310, opened 2016-12-02
Event on yaw change?


sphkress:
Hi, I'm trying to show a map overlay to make visitors show more clearly, where the selected hotspot/scene is located. The scene change can be caught by the event, that's fine so far.

The next step would be to show the selected view angle on that map. As the yaw change on a desktop is related to a 'mouseup' event, I could successfully catch the yaw angle with .getYaw. But when I'm using the keyboard, it's not working. Also, for a more fluid transition, catching both, keyup and mouseup, is always a bit late with animations. So, is there some way to catch a 'yawchange' event?

mpetroff:
If I add an event of this type, it would be a render event, called whenever a frame is rendered. I'm somewhat concerned about performance, though, from firing an event so often; I need to think about it more and do some testing.

sphkress:
I just found the floorplan feature on another viewer and there it doesn't seem to be a performance issue... but probably a different approach. Maybe this can help:
http://krpano.com/examples/116/examples/virtual-tours/scenes-with-imagemap-2/scenes-with-imagemap-2.html

strarsis:
This would be great! I currently have to track all the mouse/touch events by myself.

WaliedCheetos:
Hi There, thanks for opening the issue long time ago.

What i am looking for is when i hit (_panoramaViewer.startAutoRotate(5);) for example i want to simulate the rotation on the underlying map in order to send a real feel of panorama feeling.

So am wondering, any luck with this?

orlandobj:
For those who want to add a function for yaw, pitch or hvof changes, you can modify the panellum.js, line 1366, and add the function there. This ins't the beauty way, but will work. The event will trigg about 150 times per second and, for me, who want to make 2 views to move synchronously, this solved my problem. Note: I do not used the minified version.

strarsis:
[Event debouncing](https://gomakethings.com/debouncing-your-javascript-events/) can help here.

orlandobj:
This event is already debounce in the main library. I just used this whit the same timing Petroff does.

scottackerman:
> For those who want to add a function for yaw, pitch or hvof changes, you can modify the panellum.js, line 1366, and add the function there. This ins't the beauty way, but will work. The event will trigg about 150 times per second and, for me, who want to make 2 views to move synchronously, this solved my problem. Note: I do not used the minified version.

While not optimal, this is perfect for my prototyping! Thanks :)

fareed945:
[...]

PR #922, opened 2020-10-05
Add events for mouse/touch move

Elbandi:
It would be good if there is an event for mouse/touch move.

Possible usecase: print/display current yaw/pitch/hfov values when it changed


mpetroff:
There's nothing stopping you from adding `mousemove` and `touchmove` event listeners to the `<div>` you initialize Pannellum with, using the normal mechanism (`addEventListener`). Thus, adding these events to Pannellum is redundant.

As for your possible use case, this PR doesn't handle other input methods such as keyboard inputs. This more general type of event to handle view changes is discussed in #310, but I have some reservations about it related to rate limiting.

Thanks for the attempt, but I'm closing this PR as it only provides redundant functionality and doesn't address the nuances of a more general view change event.

PR #929, opened 2020-10-21
Fix unsafe link

aliccer:
Is it possible to fix this issue ?
https://web.dev/external-anchors-use-rel-noopener/?utm_source=lighthouse&utm_medium=devtools


mpetroff:
Thanks, but this is incomplete. There are additional places where `target="blank"` is used (actually, these should be changed to only use it in the standalone viewer, which is usually used in an `<iframe>`). I made this change in 776775f1d821b6b032a4c4175ceff20e21d66f36. As including `noreferrer` and `nofollow` is not in my best interest, only `noopener` was used.

Issue #82, opened 2015-06-18
Support for Google Cardboard


hhkaos:
I was thinking the other day that it would be great to make an app like Google Expeditions for Google Cardboad but using web technologies (p.e. mixing Pannellum + Ionic Framework).

What do you think about doing a pannellum's fork to start a Pannellum Cardboard o something like this?

Cheers!
Raul

mpetroff:
I think adding WebVR support directly to Pannellum might be a better route, as this would (eventually) allow a variety of VR headsets to be supported. From what I've read, there's currently WebVR support in Firefox Nightly for the Oculus Rift only, with plans to add Cardboard support, and there are experimental Chromium builds that support WebVR for both the Rift and Cardboard.

There's also the school of thought that VR is sufficiently different from the regular viewer and shouldn't be integrated directly as it requires a completely different UI. In this case, it would make the most sense to write a separate Pannellum VR viewer using libpannellum.

ofekd:
I've been able to use this (somewhat) successfully on Firefox for Android:
https://vr.chromeexperiments.com/

I guess it doesn't use WebVR - are you insistent on waiting for it?

Thanks for looking into it :+1:

hhkaos:
Hi guys!

I have tried both :), WebVR doesn't work very good in my phones (LG G3 and Nexus 5) but it seem to work better on my desktop. It crashes in Firefox and in Chrome very frequently.

About Chrome Experiments works great.

But I also tried:
- https://github.com/hawksley/eleVR-Picture-Player
- https://github.com/hawksley/eleVR-Web-Player

It works better than WebVR but not on every phone I think.

Thanks!

ofekd:
@mpetroff - what's your current view on this? We would like to start development and are wondering about integration vs separate viewer. In any case we will probably do something similar to Chrome Experiments.

mpetroff:
I don't feel particularly strongly about VR, so any sort of implementation isn't a priority for me, particularly since VR technology and standards have yet to stabilize. At the moment, I'd tend toward a separate viewer since I'm not sure I want to integrate something that's still in so much flux (and I haven't implemented a plugin architecture yet, which would be a happy medium between integration and a separate viewer).

ofekd:
It looks like having a plugin architecture along with a proper API like as mentioned in #30 will enable a lot of progress. Do you need any help with it? Where is the work coordinated?

mpetroff:
I just need time to work on it, which always seems to be in short supply.

[...]

PR #976, opened 2021-03-25
Stereo media support

rawnsley:
Support for stereo images and videos in top-bottom configuration. Implemented through the addition of a "vertical texture range" uniform for the shader, which applied as a factor to the y texture lookup coordinate. It defaults to 1.0 (essentially a noop), but for stereo media it is set to 0.5 so the texture range is limited to the top half of the texture.

### DIFF
--- a/src/js/libpannellum.js
+++ b/src/js/libpannellum.js
@@ -374,7 +374,7 @@ function Renderer(container) {
         program.drawInProgress = false;
 
         // Set background clear color (does not apply to cubemap/fallback image)
-        var color = params.z ? params.backgroundColor : [0, 0, 0];
+        var color = params.backgroundColor ? params.backgroundColor : [0, 0, 0];
         gl.clearColor(color[0], color[1], color[2], 1.0);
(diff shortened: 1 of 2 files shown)


mpetroff:
Thanks for taking the time to make a pull request, but I don't see a compelling reason to include this. However, I could be convinced otherwise with a good reason.

Why should the user be forced to download twice as much image data, when a non-stereo source image could be prepared instead? This functionality does not seem to be in the best interest of the end-users viewing panoramas, IMHO.

rawnsley:
I agree about efficiency, but in our case it's just convenient to use a single video for stereo and mono clients. Stereo is a pretty obscure use case though so happy for the PR to be rejected; the fork is there if anyone needs it. We very much appreciate your work.

mpetroff:
If I ever implement support for WebXR (tracked in #82) and thus stereo support in Pannellum, I'll merge this or make a similar change. Without stereo support in Pannellum, supporting stereo source images is a pretty obscure use case.

PR #980, opened 2021-04-13
Allow JSON specified hotspot function handlers

wgmyers:
Fixes issue mentioned here: https://github.com/mpetroff/pannellum/issues/252#issuecomment-281464915

My first pull request here: I have read the contributing guidelines and hope I've got everything right.

Currently there is no way to specify `createTooltipFunc` or `clickHandlerFunc` in the JSON config file, as the strings passed in are not recognised as functions, so calls to them fail.

This patch fixes the issue by explicitly searching the contents of the JSON `hotSpots` item for anything ending in `Func` and replacing the string with the relevant JS function reference.

Arguments can be passed using `createTooltipArgs`or `clickHandlerArgs` in the normal way.

NB - The JS file containing these functions must be included prior to the inclusion of standalone.js or it won't work.

### DIFF
--- a/src/standalone/standalone.js
+++ b/src/standalone/standalone.js
@@ -77,6 +77,18 @@ function parseURLParameters() {
                 if (configFromURL.hasOwnProperty(key)) {
                     continue;
                 }
+                
+                // Handle function handlers in hotspots
+                if(key === "hotSpots") {
+                    responseMap[key].forEach((arr) => {
+                        for (var hkey in arr) {
+                            if(hkey.endsWith("Func")) {
+                                arr[hkey] = window[arr[hkey]];
+                            }
+                        }
+                    });
+                }
+                
                 configFromURL[key] = responseMap[key];
             }
 


mpetroff:
Thanks, but the `standalone.js` file is not part of Pannellum's public API. It's only for internal use as part of the `pannellum.htm` standalone version.

Click handlers are only intended to be used with Pannellum's API, not with the standalone version. Obviously, you're free to adapt the standalone version to your own use, but I'm not going to merge such changes. Also, this PR breaks the `pannellum.htm` build process, since the PR uses ES6 syntax.

joshuakoh1:
So I've spent almost a day trying to figure a workaround for this issue. 

All I want is to standardize the icons for tooltip/scene change but apparently I have to create a custom function for. Which doesn't work because I get my pannellum data via API in JSON format.

Not sure why you don't want to implement a one liner fix that would use window[functionName] so it could take a string value as well.

mpetroff:
@joshuakoh1 If you want to change the icon, you can do so using the hot spot `cssClass` configuration parameter to specify a new CSS class, which specifies the new icon as the hot spot background.

This isn't a one line fix; it's non-functional bloat for the designed usage of the standalone viewer. The standalone viewer can't use a custom function without the viewer being edited. However, the standalone viewer isn't meant to be edited; for such functionality, the API is meant to be used. If you're editing the standalone viewer, you're using it as example code. In this case you're more than welcome to make your own changes, but such changes don't belong in the default standalone viewer.

PR #997, opened 2021-06-02
add %l0 placeholder for 0 based level numbering in MultiresNode

indus:
I'm using an 3rd party tool to produce the tiles for the faces - it offers 4 patterns for file and folder hirarchy/naming but all start at level 0.
This is an easy fix that may come handy for others as well.


mpetroff:
Thanks!

PR #998, opened 2021-06-08
Reduce the main thread blocking time on load scene

Quicksilver0218:
The current scene loading method blocks the main thread for so long. I've tried to modify some codes to reduce the blocking time. Here are the modifications for your reference. However, the result is still not satisfying and I think there should be better ways to achieve the purpose. So, please don't just merge the commits without further discussion.


mpetroff:
Pannellum currently uses an ES5 code base. Until such time that I decide to modernize it and break compatibility with really old browsers (which I really should do at some point), anything that uses newer functionality such as arrow functions, promises, or `fetch` won't be merged. For example, IE11 is still supported at the moment, but this PR would break that support.

PR #1015, opened 2021-07-30
Update readme.md for multires/generate.py

INF800:
Additional help added if `pip3 install --user pyshtools` fails.

### DIFF
--- a/utils/multires/readme.md
+++ b/utils/multires/readme.md
@@ -26,6 +26,14 @@ $ sudo apt install python3 python3-pil python3-numpy python3-pip hugin-tools
 $ pip3 install --user pyshtools
 ```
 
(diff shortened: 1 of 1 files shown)


mpetroff:
I'm not going to merge broken directions for your specific system. There is no `fttw` package; it's `libfftw3-dev` in Debian and Ubuntu (and there's no `fttw` package anywhere, since the library is F**F**TW).

You only need those packages if you're trying to build `pyshtools` from source. PyPI has binary wheels for x86_64 Linux, macOS, and Windows, so should only need to build from source if you're using a different, less-common architecture.

I added a generic note about consulting the [pyshtools build-from-source directions](https://shtools.github.io/SHTOOLS/python-installing.html#build-from-source) in case of an issue in 37453a99b4de5395d7d6970d6baddad36c9f8fb8.

PR #1024, opened 2021-09-06
fix: render transforms without wgbgl

F-loat:


mpetroff:
In the future, please include a description of what the issue is and what your change does to fix it. I had to test the CSS fallback renderer myself to figure out what the issue was.

In this case, your proposed change did not fix the underlying issue but just changed what was broken. An optimization introduced in 5eab5366f6987a4abdf648cbf5f7a243f14faf28 changed the cube side array so it would only need to be created once. However, different arrays were needed for WebGL-based multires rendering and the CSS fallback renderer, which I had overlooked in making the optimization. This broke the CSS fallback renderer for cubemaps. Your proposed changed fixed it in this case but instead broke the CSS fallback renderer for panoramas in the multires format. I've properly fixed the issue in 09676322402b0c3b71ea2ef20818377fa07aa338, so the CSS fallback renderer works in both cases.

PR #1055, opened 2021-12-23
add types for typescript

dbwodlf3:
I added types for typescript.
I would like to devlop in typescript level.

I know pannellum is not commonjs working. it works only on browser.

It doesn't work in commonjs but anyway i can get typehinting

### DIFF
--- a/package.json
+++ b/package.json
@@ -16,8 +16,13 @@
   "license": "MIT",
   "homepage": "https://pannellum.org/",
   "main": "build/pannellum.js",
+  "types": "types/index.d.ts",
   "keywords": [
     "panorama",
     "viewer"
-  ]
+  ],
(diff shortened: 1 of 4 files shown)


mpetroff:
The `index.d.ts` file looks like it was automatically generated. How was this done?

My tendency is to not merge this as is, since it is an additional maintenance burden for something I don't use and am not in a position to test. Furthermore, the comments are all copied into an additional file and need to be maintained in two places. If there's a straightforward way to automatically generate the types, I'd consider including such a script, the output of which could potentially be included in releases, similar to how the minified files are generated for releases. However, the lack of testing would still be an issue.

dbwodlf3:
actually it written by hand. i will find and update it (automated index.d.ts from source code).

PR #1122, opened 2022-09-27
Adding to available features within configFromUrl keys for standalone iframe

hootstheowl:
Adding the following pannellum js config keys to standalone iframe as URL params:

compass,
showFullscreenCtrl,
showZoomCtrl,
keyboardZoom,
mouseZoom,
draggable,
showControls,
~type~

### DIFF
--- a/src/standalone/standalone.js
+++ b/src/standalone/standalone.js
@@ -58,7 +58,6 @@ function parseURLParameters() {
             case 'showControls':
                 configFromURL[option] = JSON.parse(value);
                 break;
-            case 'type':
             case 'author':
             case 'title':
             case 'firstScene':
--- a/src/standalone/standalone.js
+++ b/src/standalone/standalone.js
@@ -32,16 +32,40 @@ function parseURLParameters() {
         if (value == '')
             continue; // Skip options with empty values in URL config
         switch(option) {
-            case 'hfov': case 'pitch': case 'yaw': case 'haov': case 'vaov':
-            case 'minHfov': case 'maxHfov': case 'minPitch': case 'maxPitch':
-            case 'minYaw': case 'maxYaw': case 'vOffset': case 'autoRotate':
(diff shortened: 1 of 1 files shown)


mpetroff:
Those parameters weren't left out by accident. The idea is to only have the most-basic configuration options exposed as URL parameters. You're going to have to give a compelling explanation as to why some or all of the parameters you've proposed adding should be exposed.

hootstheowl:
This was initially intended only to be a part of my fork, for a personal iframe build in my own application. I had assumed the limited parameters were chosen for _some_ reason, but figured I would put out this pull request to see if any of the added parameters could be useful for basic customization.

With the exception of **type**, I feel that the remaining params are fairly basic. They only serve to expose access to some very rudimentary interactivity, as well as simply show / hide control interfaces. I don't think these additions, with the exception of type, are all that out of the bounds of what constitutes a basic configuration. I honestly just don't see a compelling reason for why they shouldn't be exposed.

I'm not married to the idea that repo somehow requires these updates, I just found them useful in a case where I was hosting my own iframe build.

mpetroff:
I thought about this some more and decided that my main issue was the lack of consistency this PR introduced. It added `keyboardZoom`, `mouseZoom`, and `draggable` but left out `disableKeyboardCtrl` and `doubleClickZoom`, which are in the same category of options. It also left out `dragConfirm`, which is in the same category as `compass`, `showFullscreenCtrl`, `showZoomCtrl`, and `showControls`. Thus, I added these additional parameters in 2198c2599604b1eaed9830b527cad6f7091c251d after merging the PR and updated the documentation to mention all of the parameters added.

strarsis:
+1, great additions, notably the options for UX confirmation, `dragConfirm` and `mouseZoom`. When embedding a panorama quickly in a page, those options can come in very handy!

PR #1131, opened 2022-12-09
Add support of hotspot thumbnail image

Wolfshe:
The proposed modification allows to add thumbnail at each hotspot linker. The thumbnail is display when the mouse enter in the hotspot linker element.

The thumbnail image is defined in the config json file with the "thumbnail" attribute.

### DIFF
--- a/src/js/pannellum.js
+++ b/src/js/pannellum.js
@@ -1863,6 +1863,23 @@ function createHotSpot(hs) {
     else
         div.className += ' pnlm-hotspot pnlm-sprite pnlm-' + escapeHTML(hs.type);
 
+        if ( 'thumbnail' in hs ) {
+            let thumbnail = hs.thumbnail;
+            div.addEventListener("mouseenter",function(){
+            div.style.backgroundImage = `url(${thumbnail})`; 
+            div.style.border = "solid";
+            div.style.backgroundPosition = "center";
+            div.style.backgroundSize = "contain";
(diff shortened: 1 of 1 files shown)


mpetroff:
Unless I'm missing something, this can already be done by setting the `cssClass` hot spot parameter and using the `:hover` selector for the custom CSS class. Doing this styling using CSS is better for performance reasons, anyway.

Wolfshe:
The proposed feature allows to load a specific thumbnail for a hotspot (like preview rendering of the scene associated to the hotspot). This implies to use the thumbnail attribute loaded in javascript to craft adapted css attribute.

This feature cannot be implement by a custom CSS class with :hover because hotspots don't share same thumbnail.

Best regards

mpetroff:
> This feature cannot be implement by a custom CSS class with :hover because hotspots don't share same thumbnail.

Yes, it can. You just give each hot spot a different CSS class.

mpetroff:
> This feature cannot be implement by a custom CSS class with :hover because hotspots don't share same thumbnail.

Yes, it can. You just give each hot spot a different CSS class.

syifaa34:
hi, can i know how to solve the problem where i put the animation for the custom arrow hotspot in my virtual  tour and it will be positioned based on pitch and yaw that i have set in javascript file. Right now, the custom arrow hotspots are positioned at top left corner.

PR #1132, opened 2022-12-09
iphone workaround for fullscreen

the-ft:
this is a workarround for the unsupported fullscreen api on ios (iphones/ipads).
i have added a new config value "workaroundFullscreen" (by default on) which now forces to show the fullscreen button.
on click it maximizes, additionally to the fullscreen api function, the pannellum container class. so if the api is not working, you still get something like a fullscreen inside the browser. on iphones it now looks like a fullscreen.

the config value is not really needed nor really useful. it was just a switch for the workarround to check the difference. i didnt wanted to throw the whole logic upside down. please change this as you see fit.

### DIFF
--- a/src/css/pannellum.css
+++ b/src/css/pannellum.css
@@ -480,3 +480,14 @@ div.pnlm-tooltip:hover span:after {
 .pnlm-pointer {
     cursor: pointer;
 }
+
+.pnlm-workaroundfullscreen {
+    position: absolute !important ;
+    top: 0; 
+    right: 0; 
+    bottom: 0; 
+    left: 0; 
+    height: 100% !important;
+}
+
+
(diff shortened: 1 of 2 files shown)


mpetroff:
I'm not in favor of adding a kludge just because Apple can't be bother to support a widely-supported feature of the web platform. Complain to Apple about them not supporting the fullscreen API.

mpetroff:
There was nothing wrong with your pull request. It's just not something I want to include.

the-ft:
dont worry, i'm not offended 😉. your point is legit, noone likes to mess up their code. 
it s like supporting a particular kind of personality disorder by making open source available to everyone, even those with high entitlement to special treatment. who will or should give in?

PR #1133, opened 2022-12-10
Documentation: Hotspot "`click`" event types

strarsis:
The [documentation states](https://pannellum.org/documentation/reference/#:~:text=hot%20spot%E2%80%99s%20click%20event) that the hotspot `click` event is handled by the function set in `clickHandlerFunc` field.
However, not only the hotspot `click` event, [but also the `pointerup` and `touchend` events](https://github.com/mpetroff/pannellum/blob/master/src/js/pannellum.js#L1938-L1949) will cause the `clickHandlerFunc` to be called - which makes sense as just the `click` event only works with a mouse input, not with touch (as being usually the sole input method on current mobile devices).

Maybe this should rather be named "hotspot selection" instead of "hotspot click", encompassing all events/interactions that indicate the intend to select that hotspot? But renaming this field would introduce a breaking change, so instead just name it like this in the documentation?

### DIFF
--- a/doc/json-config-parameters.md
+++ b/doc/json-config-parameters.md
@@ -335,7 +335,7 @@ spot tooltip DOM instead of the default function. The contents of
 #### `clickHandlerFunc` (function) and `clickHandlerArgs` (object)
 
 If `clickHandlerFunc` is specified, this function is added as an event handler
-for the hot spot's `click` event. The event object and the contents of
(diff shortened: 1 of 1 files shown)


mpetroff:
Good suggestion! However, I'm a stickler for serial commas, and your commit didn't include one. As it was a one-line change, it was easier for me to just make a new commit in 54b7b22d03d74eec16147027f0d55ca623153021 than accept the pull request and then add the comma.

PR #1165, opened 2023-05-23
my commit

nohosh:
texture and coordinatets


mpetroff:
This commit has nothing but two useless comments. Please don't waste my time.

Issue #1193, opened 2023-12-11
Add option to show clamped offscreen hotspots


Malgalad:
I wonder if it would be reasonable to add an option to show offscreen hotspots at the edges of the screen? In a way GUI usually works in games, if an important marker is not visible in FoV it is stickied to the edge until you rotate camera to look at it. This way you know it exists in a scene and which way to rotate camera to look at it. Just googling random clip, like [here](https://youtu.be/xHtCQHL_5vE?t=65) you can see that "P" marker for platform is always on screen no matter how camera is rotated, with direction shown when camera is looking away.

From a technical perspective, I imagine it would look like an option that if hotspot(s) is not currently visible, clamp target yaw/pitch to -1/1 relative to current view...

PR #1195, opened 2023-12-18
Add "sticky" option to hot spots

Malgalad:
A barebones proposal to cover #1193. If hot spot has a new option `sticky` set to truthy value, then hot spot coordinates are clamped to always be within canvas space.

It seems to be working for me and does cover the basic functionality, but I am open to updates if it requires more configuration or logic. E.g. more padding, pointing arrows, not dispatching events unless hot spot should be visible without clamping, maybe some crowding logic so multiple hot spots don't collapse into one in a corner...

### DIFF
--- a/doc/json-config-parameters.md
+++ b/doc/json-config-parameters.md
@@ -358,6 +358,11 @@ changes in local image scale that occur due to distortions within the viewport.
 If set to a number, a fixed scaling is applied relative to the default hot spot
 size. Defaults to `false`.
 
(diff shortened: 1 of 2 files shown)


mpetroff:
This implementation is flawed. It doesn't handle hot spots that are >90 deg from the center of the view, and the vertical position of these indicators should not move when panning only in yaw. Furthermore, there would need to be some sort of visual indicator that these hot spots are actually off the edge of the screen.

PR #1205, opened 2024-02-14
Pin pyshtools to 4.9.1 in multires generator Dockerfile.

kabalin:
Latest version of pyshtools uses Meson for building and does not build (missing dependencies, bumping ubuntu and adding [listed here](https://shtools.github.io/SHTOOLS/python-installing.html#how-to-install-blas-lapack-and-fftw3) does not help, may be missing something obvious), thus proposing a quick fix to pin version and make it work again.


mpetroff:
Seems like a reasonable fix. Thanks!

mpetroff:
For future reference, it appears that `libopenblas-dev` is needed instead of `libblas-dev`, as is `pkg-config`.

PR #1216, opened 2024-05-12
Github actions for Docker image

Elbandi:
This pr do lots of thins:
- Adds a github action to build generate-panorama and push to gh container registry
- Updated the image base to latest (24.04) Ubuntu LTS.
- Add renovate bot to track and upgrade image and ubuntu package versions (this will trigger the action to rebuild image)

For renovatebot you need a PAT token (read only access).

### DIFF
--- a/.github/workflows/release-container.yml
+++ b/.github/workflows/release-container.yml
@@ -5,6 +5,9 @@ on:
   push:
     branches:
       - master
+    paths:
+      - '.github/workflows/release-container.yml'
+      - 'utils/multires/**'
     tags:
       - "*"
 
--- a/.github/workflows/release-container.yml
+++ b/.github/workflows/release-container.yml
@@ -1,0 +1,52 @@
+name: Create and publish a container image on release
+
+on:
+  workflow_dispatch:
+  push:
+    branches:
+      - master
+    tags:
+      - "*"
+
+env:
+  REGISTRY: ghcr.io
+  IMAGE_NAME: ${{ github.repository }}
+
+jobs:
+  build-and-push-image:
+    runs-on: ubuntu-latest
+    permissions:
+      contents: read
+      packages: write
+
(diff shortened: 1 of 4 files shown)


mpetroff:
This "fixes" things that aren't broken.

1. I have no desire to deal with the Github container registry (or any other container registry).
2. It's far too soon to update to Ubuntu 24.04. I'd at least wait for 24.04.1. The image base should be updated within the next year, though, before 20.04 goes out of support.
3. I see no benefit to setting explicit version numbers in the Dockerfile. And then adding a third-party service to create pull requests to update these explicit versions would just waste my time with reviewing unnecessary pull requests.

I appreciate the effort, but I don't see any value in these changes.

Elbandi:
1. ok, as you want.
2. [ubuntu focal](https://endoflife.date/ubuntu) end in 10 months. and i dont know why, [hugin-tools is missing](https://packages.ubuntu.com/search?keywords=hugin-tools) from 22.04 LTS. 
3. security upgrade. If someone dont want to use 1-2 years old image, this is a possible solution to rebuild docker container if needs.

mpetroff:
Hugin was missing from Debian unstable for a brief period of time in early 2022 due to a dependency issue. It just happened to overlap with the time when Ubuntu 22.04 was branched off from it. The base image can be updated in a few months, once the bugs are worked out of 24.04 but still well before 20.04 goes out of support.

If you don't pin version numbers, the latest versions will be used. Pinning version numbers forces older packages to be used until the pinned version numbers are updated, which is the potential security issue, not the other way around.

mpetroff:
> So you dont have to watch packages every day/weeks and start imagebuild yourself

Currently, I don't need to watch anything. With this change, I would need to deal with useless pull requests. Thus, it would make more work for me. Thus, it isn't happening.

PR #1238, opened 2024-09-25
Add missing callback invocations to setPitch, setYaw and setHfov.

jox:
**Issue:**
The methods `setPitch`, `setYaw`, and `setHfov` accept a callback function as an argument, but there are cases where this callback is not invoked. This behavior is unexpected and likely represents a bug.

**Expected Behavior:**
The callback should be invoked consistently whenever the methods are called, regardless of specific conditions.

**Observed Behavior:**
The callback **is invoked** under the following conditions:
1. When the `animate` argument is `true`.
2. When the change in value is smaller than a defined `epsilon`, regardless of the value of `animate`.

The callback **is not invoked** under the following condition:
- When `animate` is `false` **and** the change in value is greater than `epsilon`.

**Impact:**
While the callback is documented to be triggered "when animation finishes", it would be illogical to skip the callback entirely when no animation is requested (`animate = false`). Even without animation, the callback should still be called once the value has been set, as this is essential for proper handling of the method's completion state. The current behavior of skipping the callback in specific non-animated cases is unexpected and likely unintended by design.

### DIFF
--- a/src/js/pannellum.js
+++ b/src/js/pannellum.js
@@ -2721,6 +2721,8 @@ this.setPitch = function(pitch, animated, callback, callbackArgs) {
             setTimeout(function(){callback(callbackArgs);}, animated);
     } else {
         config.pitch = pitch;
+        if (typeof callback == 'function')
+            callback(callbackArgs);
     }
     animateInit();
     return this;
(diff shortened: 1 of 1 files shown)


mpetroff:
> While the callback is documented to be triggered "when animation finishes", it would be illogical to skip the callback entirely when no animation is requested

I disagree and think that it's perfectly logical. Without animation, there's no reason to have a callback at all as the callback function could just be executed immediately after the call to `setPitch` / `setYaw` / `setHfov` by the code calling those functions.

> The current behavior of skipping the callback in specific non-animated cases is unexpected and likely unintended by design.

As there is no reason for a callback without animation, there was also no reason to check for whether there was animation for the epsilon case. Calling the callback in a specific non-animated case was intentional, to avoid an unnecessary check.

While the current behavior is intentional, I can see how it could cause confusion, so your change is reasonable.

PR #1240, opened 2024-10-03
add support for different panorama in fullscreen

zbycz:
Hi,

first, thanks for an excellent panorama viewer :-) On my opensource map browser OsmAPP I use pannellum to display street-view panoramas. See eg https://osmapp.org/50.07749,14.44896

Though, currently pannellum accepts only one equirectangular panorama, which is usually several megabytes to load and user scarcely opens it. My idea is to load smaller `panorama` in standard embedded view, and only switch to `panoramaFullscreen` when changing to the fullscreen mode.

Do you think it could be mergable? Or perhaps is there another solution without modyfing pannellum's code?

Thanks!

### DIFF
--- a/examples/example-fullscreenPanorama.htm
+++ b/examples/example-fullscreenPanorama.htm
@@ -1,0 +1,10 @@
+<!DOCTYPE HTML>
+<html>
+<head>
+    <title>pannellum embed example</title>
+    <meta http-equiv="Content-Type" content="text/html; charset=UTF-8" />
+</head>
+<body>
+    <iframe width="480" height="390" allowfullscreen style="border-style:none;" src="../src/standalone/pannellum.htm#panorama=../../examples/examplepano-640.jpg&amp;panoramaFullscreen=../../examples/examplepano.jpg&amp;autoLoad=true&amp;showZoomCtrl=false&amp;autoRotate=-1"></iframe>
+</body>
+</html>
--- a/src/js/pannellum.js
+++ b/src/js/pannellum.js
@@ -2327,6 +2327,21 @@ function processOptions(isPreview) {
 function toggleFullscreen() {
     if (loaded && !error) {
         if (!fullscreenActive) {
+            if (config.panoramaFullscreen) {
+                var p = '';
+                if (config.basePath) {
+                    p = config.basePath;
+                }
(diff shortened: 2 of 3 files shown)


mpetroff:
> Do you think it could be mergable?

No, it isn't. Using the multires format is the intended way to dynamically handle different sizes and zoom levels without wasting bandwidth and increasing load times. I assume you're not using it because you already have equirectangular images and don't want to generate tile sets for each one.

> Or perhaps is there another solution without modyfing pannellum's code?

This is pretty straightforward to implement using two scenes and an event handler. Something like this:
```html
<!DOCTYPE HTML>
<html>
<head>
    <meta charset="utf-8">
    <meta name="viewport" content="width=device-width, initial-scale=1.0">
    <title>Switch on fullscreen</title>
    <link rel="stylesheet" href="https://cdn.jsdelivr.net/npm/pannellum@2.5.6/build/pannellum.css"/>
    <script type="text/javascript" src="https://cdn.jsdelivr.net/npm/pannellum@2.5.6/build/pannellum.js"></script>
    <style>
    #panorama {
        width: 600px;
        height: 400px;
    }
    </style>
</head>
<body>

<div id="panorama"></div>
<script>
var viewer = pannellum.viewer('panorama', {
    "default": {
        "firstScene": "small",
        "autoLoad": true
    },
    "scenes": {
        "small": {
            "panorama": "small.jpg",
        },
        "big": {
            "panorama": "big.jpg",
        },
    }
});
viewer.on('fullscreenchange', function(fullscreenActive) {
    var pitch = viewer.getPitch(),
        yaw = viewer.getYaw(),
        hfov = viewer.getHfov();
        scene = fullscreenActive ? 'big' : 'small';
    viewer.loadScene(scene, pitch, yaw, hfov);
});
</script>

</body>
</html>
```

zbycz:
> I assume you're not using it because you already have equirectangular images and don't want to generate tile sets for each one.

Yes, you are exactly right. Since I consume a 3rd party API, i cannot generate it on my own.

> This is pretty straightforward to implement using two scenes and an event handler. Something like this

Wow, thank you very much for the code. ❤️ I have already [implemented it](https://github.com/zbycz/osmapp/pull/628) and deployed to production. 

I must say, I didn't realized that scenes could be used to load more equirectangular panoramas. It could be useful if you added an example like this to your documentaion. An example like "Changing scenes programatically" would be enough for me to find out solution. I already experimented with the event, but was unable to change the image :-)

PR #1260, opened 2025-06-06
Update orientationYawOffset on scene change (fixes #419)

### LINKED ISSUE #419: Orientation and sameAzimuth bug?
When I am navigating my tour with sameAzimuth set, each new scene uses the correct azimuth unless I also have the orientation control turned on. With orientation turned on, the compass direction jumps with each scene load. I'm wondering if it is not taking the northOffset into account? If I am dragging the camera around with touch or with a mouse then it works fine, there's only a problem when orientation is turned on and I'm looking around by moving the phone or tablet. Tested on both Android and Windows (with a surface tablet) in Chrome.

robertaugusteijn:
When using device orientation for pointing control and loading a new scene, the new config.yaw was not taken into account. This may cause undesired jumps in viewing direction.

This change triggers recalculation of the orientationYawOffset with the config.yaw of the newly loaded scene.

### DIFF
--- a/src/js/pannellum.js
+++ b/src/js/pannellum.js
@@ -2554,6 +2554,10 @@ function loadScene(sceneId, targetPitch, targetYaw, targetHfov, fadeDone) {
     if (workingHfov !== undefined) {
         config.hfov = workingHfov;
     }
+    
+    // Trigger recalculation of orientationYawOffset in orientationListener()
+    if (orientation === true) orientation = 10;
+	
     fireEvent('scenechange', sceneId);
     load();
 }


mpetroff:
Thanks! Just to confirm, since it wasn't entirely clear in the original issue, this fix only applies when `northOffset` is used, correct? At least that's the only bug I was able to reproduce that this change fixed.

robertaugusteijn:
Thanks for merging!

I guess this fix is needed for all yaw affecting config:
- `yaw` in default/default scene/scene/initial config
- `targetYaw` (number / `sameAzimuth`)

Only for `targetYaw=same` the recalculation will result in the same `orientationYawOffset` and will have no visible effect.

Also while testing this, I discovered that using `orientationOnByDefault=true` will act as a workaround. Because `startOrientation()` will then be called on each scene reload which sets `orientation` to `1`, triggering a recalculation of `orientationYawOffset`. At least I could confirm this for the latest release (2.5.6).

